PR #1, opened 2016-09-07
:book: Add placeholder readme

neilbmclaughlin:
Created to force Heroku review app


st3v3nhunt:
Does this need merging? And is it just the changes in `app.json` that need to be included?

PR #18, opened 2016-09-13
:mega: Refactored to middleware structure

st3v3nhunt:


st3v3nhunt:
The output from the route is going to change along with the way the rendering of it is done i.e. view vs messages as this progresses.

The tests are quite similar atm. I think over time the server level tests will be changed to accommodate the changing structure and output of the application. The location validator is going to change as functionality is added e.g. places of interest, towns, etc. but the tests for the postcode validation are likely to be required in a form closely related to how they currently are. Which is a long way of saying I know the tests might look like they are copied but I think that is OK atm!

PR #20, opened 2016-09-14
:spades: Extract environment stuff

st3v3nhunt:


config/environment.js
```diff
@@ -0,0 +1,14 @@
+const dotenv = require('dotenv');
+const requireEnv = require('require-environment-variables');
+
+function validate() {
```
neilbmclaughlin: Strictly speaking this is configuring (the 2nd part is configuring validation).

Bit pedantic but then so is lint :)

Also, giving the new GitHub review process a spin.

st3v3nhunt: So it is. What change(s) do you suggest?

neilbmclaughlin: Rename to configure()?

[Note: had to do a manual PR page refresh for your comment to appear. Was aware of your comment through Slack]

st3v3nhunt: Done

PR #22, opened 2016-09-15
Feature/process results

st3v3nhunt:
This is not the finished thing for showing only open results (far from it), however, it is getting the data from the api, processing it and returning the top 3 results. Next stuff includes:

* Filtering by only open (using the moments lib)
* Adding views
* Extra error handling

However, this should be ok for a review with the idea to get it merged in to `master` sooner rather than later.


middleware/getResults.js
```diff
@@ -0,0 +1,41 @@
+const http = require('http');
+const async = require('async');
+const xmlParser = require('../lib/xmlParser');
+const utils = require('../lib/resultsUtils');
+const processResults = require('../lib/resultsProcessor.js');
+
+function getData(requestUrl, page, cb) {
```
neilbmclaughlin: Are we adopting a convention of a small number of common abbreviations of which cb for callback will be one. I mention this as we discourage undue terseness in naming in out contribution notes.

middleware/getResults.js
```diff
@@ -0,0 +1,41 @@
...
+    res.on('end', () => {
+      // TODO: Handle responses that are not 200
+      const pharmacyList = xmlParser(xml);
+      cb(null, pharmacyList);
+    });
+  });
+}
+
+function getResults(req, res, next) {
+  const requestUrl = utils.generateRequestUrl(req);
```
neilbmclaughlin: What is the rational for not having this a middleware?

middleware/getResults.js
```diff
@@ -0,0 +1,41 @@
...
+
+  const tenRequests = [];
+  [1, 2, 3, 4, 5, 6, 7, 8, 9, 10].forEach((item) => {
+    tenRequests.push((cb) => {
+      getData(requestUrl, item, cb);
+    });
+  });
+
+  async.parallel(tenRequests,
```
neilbmclaughlin: Nice!

middleware/getResults.js
st3v3nhunt: I was wondering exactly the same when I reviewed it myself. Should `cb` be the slightly longer `callback` or is it a known convention?

I'm happy with either and would probably prefer the latter to make it clearer. What do you think?

middleware/getResults.js
```diff
@@ -0,0 +1,41 @@
...
+      cb(null, pharmacyList);
+    });
+  });
+}
+
+function getResults(req, res, next) {
+  const requestUrl = utils.generateRequestUrl(req);
+
+  const tenRequests = [];
+  [1, 2, 3, 4, 5, 6, 7, 8, 9, 10].forEach((item) => {
```
st3v3nhunt: I'm pretty sure there is a better way to generate an array of 10 functions but this works and is reasonably clear.

middleware/getResults.js
st3v3nhunt: Not much really, I was thinking about it but decided against it mainly for the reason of it feels less like a process the request needs to do through and more of a function something else relies on.

Issue #38, opened 2016-09-22
Sort out css/scss


st3v3nhunt:

st3v3nhunt:
Given the repo now has generated CSS this can be closed.

PR #49, opened 2016-09-26
:art: Integrate latest designs

st3v3nhunt:
Quite a few changes here but they are all to do with getting the latest designs integrated into the application. With one exception. I've updated the error view and made the error message hidden. It can be shown if the content of the page is clicked on.

There are files that need to be deleted which I'll either tag onto this or do as a separate PR.

A number of issues have been raised whilst I've been working on this, things like sorting out the SCSS (#38) as it is just a copy/paste of the generated CSS atm. Please add anything other issue that need fixing during the review.


neilbmclaughlin:
Not sure about the value of the displaying errors in the error page. It's really a hack for the lack of logging. As it stands if we are in development then we can see the errors in console. Same applies to preview/staging on Heroku (and I presume Azure too).

st3v3nhunt:
No doubt the error page will change in the coming days.

Review will be tomorrow when Si is back in. The results page(s) have not been updated yet.

PR #52, opened 2016-09-26
:accept: Make cookie message dynamic

### LINKED ISSUE #50: Make cookies banner disappear as per existing Beta cookie banner
Basically it just needs to have the existing JS used on the Beta added.

The functionality is to only show the message once every 30 days.

st3v3nhunt:
This change adds the functionality to have the cookie message be dynamic and inline with how it works on the beta site.

Additional work required:

* Have `nhsuk.bundle.js` not be committed to source control and have it generated as part of the deployment
* Add ability to have the file minimised for deployments in non-dev environments

This has been facilitated by https://www.npmjs.com/package/webpack which seems to be the 'better/more popular' version of things like https://www.npmjs.com/package/browserify and https://www.npmjs.com/package/requirejs

Fixes #50

### DIFF
--- a/assets/javascripts/modules/cookie-message.js
+++ b/assets/javascripts/modules/cookie-message.js
@@ -1,0 +1,44 @@
+/* global document */
+function setCookie(name, value, options = {}) {
+  let cookieString = `${name}=${value}; path=/`;
+
+  if (options.days) {
+    const date = new Date();
+    date.setTime(date.getTime() + (options.days * 24 * 60 * 60 * 1000));
+    cookieString = `${cookieString}; expires=${date.toGMTString()}`;
+  }
+
+  if (document.location.protocol === 'https:') {
+    cookieString = `${cookieString}; Secure`;
+  }
+
+  document.cookie = cookieString;
+}
+
+function getCookie(name) {
+  const nameEQ = `${name}=`;
+  const cookies = document.cookie.split(';');
+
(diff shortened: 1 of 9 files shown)


st3v3nhunt:
I'm going to remove the webpack dependency for the time being - until such as time it is used for generating the files when they change.

As far as bitHound code checks failing i'm not sure what to do about that given the file has been ignored in `.eslintignore`. What does bitHound code checks do that we don't get from other things we have running e.g. eslint?

assets/javascripts/modules/cookie-message.js
```diff
@@ -0,0 +1,44 @@
...
+function setCookie(name, value, options = {}) {
+  let cookieString = `${name}=${value}; path=/`;
+
+  if (options.days) {
+    const date = new Date();
+    date.setTime(date.getTime() + (options.days * 24 * 60 * 60 * 1000));
+    cookieString = `${cookieString}; expires=${date.toGMTString()}`;
+  }
+
+  if (document.location.protocol === 'https:') {
```
neilbmclaughlin: I share your concerns about cutting and pasting code. I can see what is going on but I'm not sure if this is good practice or not. I guess I'd opt for a minimal cookie package with testing and high usage but this comes with a cost of code bloat (e.g. https://github.com/js-cookie/js-cookie).

PR #61, opened 2016-09-27
:cherry_blossom: Display error message for invalid postcode

st3v3nhunt:


st3v3nhunt:
I have no idea why Codacy has failed the review. Maybe some of the settings need tweaking?
https://www.codacy.com/app/nhsuk/connecting-to-services/settings/quality or the required check to be removed.

app/views/search.nunjucks
```diff
@@ -15,6 +15,9 @@
         <!-- label -->
       </label>
 
+      {% if errorMessage %}
```
neilbmclaughlin: Dreaded back button behaviour:

Enter invalid/missing postcode -> error message displayed -> enter valid postcode -> results listed -> hit browser back button -> valid postcode displayed in textbox + error message displayed.

neilbmclaughlin:
Tried to look at the failing codacy check but it seems the rebase has changed the status to passing. There doesn't seem to be a time series history of checks - it is mostly about current quality. There are per commit measures on the dashboard but since the push would have been forced then it isn't much help.

app/views/search.nunjucks
st3v3nhunt: Good spot, I've fixed it of sorts. The input will not be reloaded but getting rid of the error message is another thing altogether.
Notes on how WebKit caches pages https://webkit.org/blog/427/webkit-page-cache-i-the-basics/

st3v3nhunt:
Re: the Codacy fail - not sure what that was about and as you say, it is no longer failing. Something to keep an eye on I think.

app/views/search.nunjucks
neilbmclaughlin: Are you proposing that this PR gets merged and the error message persisting is raised as an issue?

st3v3nhunt: AFIAK this is the way the browser works and until such a time that JS functionality is added to fix the issue on the front end there isn't much to be done about it. And I'm not inclined to do that as a priority now.
Happy to hear about alternatives.

st3v3nhunt: #65

Issue #62, opened 2016-09-28
Validate location for both results routes


st3v3nhunt:
At the moment there is no validation on the `/results` route. The input needs to be validated in the same way as `/results-open`

st3v3nhunt:
See #66

PR #66, opened 2016-09-28
:steam_locomotive: Use single route for all results

st3v3nhunt:
Got rid of the 2 results routes in favour or a single one.

Started to move to use `res.locals` rather than `req.<thing>`

Next iteration of `services.js` is to move logic out and into libs


config/routes.js
```diff
@@ -28,17 +28,8 @@ router.get('/search',
...
-  validateLocation,
-  urlUtils.urlForPharmacyPostcodeSearch,
-  servicesMiddleware.getPharmacies,
-  servicesMiddleware.getPharmacyOpeningTimes,
-  servicesMiddleware.prepareOpenThingsForRender,
-  servicesMiddleware.renderServiceResults
-);
-
 router.get('/results',
+  validateLocation,
```
neilbmclaughlin: I'm still able to navigate to http://localhost:3000/results-open by virtue of the `route.get('/:view',` line in config/routes. results-open.nunjuck needs to be deleted.

st3v3nhunt:
Fixes #62

app/middleware/services.js
```diff
@@ -152,34 +153,43 @@ function getDisplayValuesMapper(location) {
...
+    altResultsMessage = 'See only open places nearby';
+
+    const tenClosestPlaces = req.pharmacyList
+          .sort(sortByDistance)
+          .slice(0, 10);
+
+    serviceList =
+      pharmacyMapper(tenClosestPlaces)
+        .map(getDisplayValuesMapper(location))
+        .sort(sortByDistanceInKms);
```
st3v3nhunt: TBH, I'd not looked at the detail in this change, I was focussing on moving stuff around to get rid of the 2 methods. Although `prepareOpenThingsForRender` had only a single sort `prepareForRender` still had 2 which is why this branch has...

Is it still required? Not that I can see.

Do you want it removing as part of this or the next refactor?

config/routes.js
st3v3nhunt: I'll get that gone

app/middleware/services.js
neilbmclaughlin: It's easier to remove it than log it in an issue.

PR #68, opened 2016-09-28
:birthday: Don't run tests twice on CI

st3v3nhunt:
I've also made the githooks do the same as CI in so far as generating coverage which in turn runs the tests and checking the coverage.

Hopefully this will also help to prevent some of the build failures we have seen. Should also speedup the builds.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,6 @@ node_js:
 - node
 script:
 - npm run lint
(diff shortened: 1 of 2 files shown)


st3v3nhunt:
OK, so not having the build fail due to timeouts isn't going to happen given this build just failed for that very reason :cry:

PR #70, opened 2016-09-28
:rocket: Route for no rewriting

### LINKED ISSUE #69: Make app routing work for release
The application should handle the routing in the way it will be presented to users rather than relying on lots of rewriting rules.

st3v3nhunt:
Setup the app to minimise the amount of route rewriting required. Just at the moment this means the 'home' page is `/symptoms/stomach-ache` rather than `/`. I've specifically not made the routes take the symptom as a variable as that can be done as the next symptom is added. And symptom is likely going to need to be variable as it could also be condition.

Fixes #69

### DIFF
--- a/app/middleware/services.js
+++ b/app/middleware/services.js
@@ -108,8 +108,7 @@ function getPharmacyOpeningTimes(req, res, next) {
 }
 
 function renderServiceResults(req, res) {
-  const path = req.path.substring(1);
-  res.render(path);
+  res.render('results');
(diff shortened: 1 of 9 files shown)


st3v3nhunt:
The able param determines whether somebody is able to get there by themselves or not and changes the view rendered.

What is the issue about it not being persisted?

st3v3nhunt:
Rebased to fix merge conflicts

PR #71, opened 2016-09-28
:page_with_curl: Fix path to font CDN

### LINKED ISSUE #59: Sort out CDN font path
Load a page and check the console for errors. A number of resources are not available i.e. not found due to the path in the `.css` file not having the variable that is used in the `SCCS` replaced.

This needs sorting so that the font files are being mapped to the correct CDN host.

st3v3nhunt:
Fixes #59

### DIFF
--- a/public/css/nhsuk-ie6.css
+++ b/public/css/nhsuk-ie6.css
@@ -1,28 +1,28 @@
 @charset "UTF-8";
 @font-face {
   font-family: "FS Me";
-  src: url("/* @echo FONT_CDN_PATH */assets/fonts/fsme/FSMeWeb-Light.eot?") format("eot"), url("/* @echo FONT_CDN_PATH */assets/fonts/fsme/FSMeWeb-Light.woff") format("woff");
+  src: url("https://dhrlmnmyf2njb.cloudfront.net/assets/fonts/fsme/FSMeWeb-Light.eot?") format("eot"), url("https://dhrlmnmyf2njb.cloudfront.net/assets/fonts/fsme/FSMeWeb-Light.woff") format("woff");
   font-weight: 200;
   font-style: normal;
 }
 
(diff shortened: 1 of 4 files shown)


public/css/nhsuk-ie6.css
```diff
@@ -1,28 +1,28 @@
 @charset "UTF-8";
 @font-face {
   font-family: "FS Me";
-  src: url("/* @echo FONT_CDN_PATH */assets/fonts/fsme/FSMeWeb-Light.eot?") format("eot"), url("/* @echo FONT_CDN_PATH */assets/fonts/fsme/FSMeWeb-Light.woff") format("woff");
+  src: url("https://dhrlmnmyf2njb.cloudfront.net/assets/fonts/fsme/FSMeWeb-Light.eot?") format("eot"), url("https://dhrlmnmyf2njb.cloudfront.net/assets/fonts/fsme/FSMeWeb-Light.woff") format("woff");
```
neilbmclaughlin: This seems at odds with the outcome of the slack conversation but I'm sure I've misinterpreted it:

> Yep. We only have a license for certain domains
> 
> [5:56]  
> That domain is just our cloudfront cdn which sits in front of an s3 bucket
> 
> [5:58]  
> Also, it's best not to reference assets on beta as they're all revved and sitting on a cdn
> 
> [5:58]  
> If you need compiled assets take a copy and serve locally

st3v3nhunt: How so?

neilbmclaughlin: Reading the thread I expected them to be served from something like /fonts (i.e. a copy we hold similar to how we handle css).

st3v3nhunt: Assets refers to js, css, not for the fonts

PR #77, opened 2016-09-29
:feet: Add analytics tracking code

neilbmclaughlin:
Analytics ID's need to be populated in env vars for the client code to
be added.

Changed config so it is the same for each environment and differences
are managed through env vars (in line with 12Factor principles)

Fixed eslint warning regarding function name missing

To be considered before merge:

- Content security policy (e.g. https://github.com/helmetjs/csp)
- Assets from CDN
- Webpack usage


app/middleware/locals.js
```diff
@@ -0,0 +1,11 @@
+module.exports = (config) =>
+  (req, res, next) => {
+    /* eslint-disable no-param-reassign */
+    res.locals.GOOGLE_ANALYTICS_TRACKING_ID = config.googleAnalyticsId;
+    res.locals.WEBTRENDS_TRACKING_ID = config.webtrendsId;
+    res.locals.HOTJAR_TRACKING_ID = config.hotjarId;
+    res.locals.FONT_CDN_HOST = config.fontCdnHost;
```
st3v3nhunt: This isn't mentioned in the `README.md`

.bithoundrc
```diff
@@ -42,17 +42,13 @@
     "**/extjs/src/**",
     "**/extjs/welcome/**",
     "bower_components/**",
-    "**/assets/**"
+    "**/assets/**",
```
st3v3nhunt: Is this needed? What goes in `**/assets/**`?

.bithoundrc
```diff
@@ -42,17 +42,13 @@
...
+    "public/js/tracking/**"
```
st3v3nhunt: Should this be a bit more wide ranging? Given that the dir will have generated assets in? Or maybe wait until that time i.e. when the assets are being generated?

app/middleware/locals.js
neilbmclaughlin: Good, spot toyed with adding it but decided it was out of scope and left in in error.

.bithoundrc
neilbmclaughlin: Nothing anymore so can be removed - will do this.

.bithoundrc
neilbmclaughlin: Wait until that time.

neilbmclaughlin: webtrends.js replaced by webtrends.min.js

neilbmclaughlin: webtrends code moved into a public/js/vendor folder so we can be specific about not doing code analysis on 3rd party code.

.bithoundrc
```diff
@@ -42,17 +42,12 @@
...
     "**/tests/**",
     "**/spec/**",
     "**/specs/**"
-  ],
-  "dependencies": {
-    "unused-ignores": [
-      "webpack"
-    ]
-  }
+  ]
```
neilbmclaughlin: This trailing comma was causing bithound excludes to fail.

neilbmclaughlin:
Suggested changes made - this is good to go.

README.md
```diff
@@ -39,6 +39,18 @@ is used to check all are present as part of the application start-up. If
...
 
+Environment variables are used to set application level settings for each
+environment.
+
+| Variable                        | Description                                                               | Default                  | Required |
+|:--------------------------------|:--------------------------------------------------------------------------|:-------------------------|----------|
+| `NODE_ENV`                      | node environment                                                          | development              | | 
+| `PORT`                          | server port                                                               | 3000                     | | 
+| `GOOGLE_ANALYTICS_TRACKING_ID`  | [Google Analytics](https://www.google.co.uk/analytics) property id        |                          | |
+| `WEBTRENDS_TRACKING_ID`         | [Webtrends](https://www.webtrends.com/) tracking id                       |                          | |
```
st3v3nhunt: In `config/config.js` this is `WEBTRENDS_ANALYTICS_TRACKING_ID`
and the hotJar one is `HOTJAR_ANALYTICS_TRACKING_ID`

Not sure if you want to change it here or in config?

neilbmclaughlin: Good spot! I'll change the readme.

st3v3nhunt: Ha yeah. I was going to mention this in the first place...and the possibility of using the require env vars package but as these aren't required that ins't going to work.

app/views/layout.nunjucks
```diff
@@ -24,6 +24,21 @@
     <!--[if IE 8]><link href="./css/nhsuk-ie8.css" media="screen" rel="stylesheet" type="text/css" /><![endif]-->
 
     <link rel="stylesheet" href="./css/print.css" media="print" type="text/css">
+
+    {% if HOTJAR_TRACKING_ID %}
+       <!-- rational for this code being in head: https://docs.hotjar.com/v1.0/docs/manual -->
+      <script>
+        (function(h,o,t,j,a,r){
+        h.hj=h.hj||function(){(h.hj.q=h.hj.q||[]).push(arguments)};
+        h._hjSettings={hjid:{{ HOTJAR_TRACKING_ID }},hjsv:5};
```
st3v3nhunt: The var needs escaping e.g.
`h._hjSettings={hjid:'{{ HOTJAR_TRACKING_ID }}',hjsv:5};`

st3v3nhunt: No it doesn't - See this https://docs.hotjar.com/v1.0/docs/hotjar-tracking-code
(which I would have seen if I've read the comment!).
Hotjar IDs are numeric or at least need to be if they are to work but that is something provided by Hotjar...

st3v3nhunt:
Longest PR EVER 😜

PR #81, opened 2016-10-03
Feature/three open

### LINKED ISSUE #75: Trim location input to avoid trailing spaces causing invalid postcode issues

st3v3nhunt:
Fixes #75

### DIFF
--- a/app/lib/locationValidator.js
+++ b/app/lib/locationValidator.js
@@ -2,20 +2,22 @@ const Postcode = require('postcode');
 
 function validateLocation(location) {
   let errorMessage = null;
+  let locationToReturn = location;
 
   if (!location) {
     errorMessage = 'A valid postcode is required to progress';
   } else {
-    const postcode = new Postcode(location);
+    locationToReturn = location.trim();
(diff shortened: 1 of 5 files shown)


neilbmclaughlin:
Will review after rebase.

st3v3nhunt:
@neilbmclaughlin Rebased

app/lib/locationValidator.js
```diff
@@ -2,20 +2,22 @@ const Postcode = require('postcode');
...
-      errorMessage = `${location} is not a valid postcode, please try again`;
+    if (!postcode.valid() && !Postcode.validOutcode(locationToReturn)) {
+      errorMessage = `${locationToReturn} is not a valid postcode, please try again`;
     }
   }
 
   return {
     errorMessage,
-    input: location,
+    input: locationToReturn,
```
neilbmclaughlin: I can see why you need to do the let in line 5 to prevent side effects from references. Would it be better to make the intent explicit through a function?

e.g.  `function cloneString(string) { return cloneString = string; )`
...
`input: cloneString(location)`

PR #88, opened 2016-10-04
:moyai: Make asset routes more flexible

st3v3nhunt:


st3v3nhunt:
@neilbmclaughlin It was the action on the view causing the problems, going to a route that wasn't known!

st3v3nhunt:
Closing as #91 covers this and #90

PR #91, opened 2016-10-04
Feature/design and routes

st3v3nhunt:
Combination of the design changes and the routes as it is all needed to get the assets to be served correctly.


st3v3nhunt:
Superseded by #92

PR #98, opened 2016-10-06
:facepunch: Prevent future silent BitHound failure

neilbmclaughlin:
BitHound failed to ignore vendor files as it was silently borking
with a non valid json .bithoundrc file. Add this to prevent this
happening again.

The convoluted cat command is due to this: https://github.com/eslint/eslint/issues/4829

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
   },
   "scripts": {
     "start": "node app.js",
-    "lint": "eslint --ext .js,.json .",
+    "lint": "eslint --ext .js,.json . && cat .bithoundrc | eslint --stdin --stdin-filename=f.json",
     "test": "mocha --recursive test",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -8,7 +8,7 @@
   },
   "scripts": {
     "start": "node app.js",
-    "lint": "eslint --ext .js,.json .",
+    "lint": "eslint --ext .js,.json . && cat .bithoundrc | eslint --stdin --stdin-filename=f.json",
```
st3v3nhunt: Probably not something to worry about but this is now, AFAIK, not going to work in a Windows environment.

st3v3nhunt: Discussed this and agreed it wasn't important. If it does become an issue we can solve at that time.

PR #103, opened 2016-10-12
:oncoming_automobile: Set autofocus for usability

neilbmclaughlin:
Spoke with Michael and he's OK with this change.


st3v3nhunt:
@mcheung-nhs Do you want to do the merging on this?

mcheung-nhs:
No major accessibility issues as the main purpose of this screen is to use the input field. For browsers that support autofocus, it will save a mouse click or keyboard tabs. If more content gets added to this screen, then this may need to be revisited, but as it is now, am happy to merge.

PR #105, opened 2016-10-14
Feature/postcode lookup

st3v3nhunt:
Mostly this adds the lookup of postcodes and outcodes via https://postcodes.io for use on the file based results.

It should be OK to merge into master from the POV of not affecting what is already there. There are still things to do, as noted in the code with `TODO` comments.

Of minor note is the change to `getSampleResponse` to make it support more than just the `syndication_responses` dir.


st3v3nhunt:
Things still todo:

* Add link for Google maps
* Add link for alt results page
* Add open message
* Add open now filtering
* Remove syndication calling code

PR #111, opened 2016-10-17
:high_brightness: Added only open results

st3v3nhunt:
Using the latest branch from `moment-opening-times` e.g. git://github.com/nhsuk/moment-opening-times.git#feature/change-opening-times-format added the functionality to show open results on the results-file route.

I've made some changes to the openingTimes parser to ensure all existing functionality on the main results route continues to work.

The addition of checking the orgs for whether they are open or not has increased the duration of the filter operation. I'll look at reducing that next...

### DIFF
--- a/app/lib/getPharmacies.js
+++ b/app/lib/getPharmacies.js
@@ -1,3 +1,5 @@
+const OpeningTimes = require('moment-opening-times');
+const moment = require('moment');
 const geolib = require('geolib');
 const assert = require('assert');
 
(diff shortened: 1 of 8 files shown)


app/lib/getPharmacies.js
```diff
@@ -41,7 +47,41 @@ function nearby(searchPoint, geo, limit) {
...
+    if (openingTimes) {
+      const openingTimesMoment = new OpeningTimes(item.openingTimes.general, 'Europe/London');
+      openingTimesMessage = openingTimesMoment.getOpeningHoursMessage(moment());
+      isOpen = openingTimesMoment.isOpen(moment());
+    } else {
+      openingTimesMessage = 'Call for opening times';
+      isOpen = false;
+    }
+    /* eslint-disable no-param-reassign */
+    item.openingTimesMessage = openingTimesMessage;
```
neilbmclaughlin: I had imagined that the openingTimesMoment would be assigned to item for use by the view (e.g. item.openingTimesMoment = openingTimesMoment).

st3v3nhunt: I did have that but figured there was no need and probably didn't want to have anything that wasn't necessary being passed along

neilbmclaughlin: There seems to be some duplication going on. In the services middleware getDisplayValuesMapper is handling the population of the opening times message. There probably is no need for getDisplayValuesMapper now but some of the functionality needs moving to here.

st3v3nhunt:
Regarding the duplication - I intend to remove it at the same time as getting rid of the syndication calls

PR #114, opened 2016-10-17
:mailbox_closed: Handle not found postcode from lookup

### LINKED ISSUE #80: Handle not found location
This is effectively handling the scenario when there are no results for a given location. There is a difference between a valid postcode and one that isn't known. However, the current API being used doesn't differentiate this.

st3v3nhunt:
This handles an unknown postcode from the lookup with postcode.io

Tests in place to ensure the correct view is returned.

I've extracted the message into its' own module as it was being duplicated in a number of places.

Fixes #80

### DIFF
--- a/app/lib/locationValidator.js
+++ b/app/lib/locationValidator.js
@@ -1,4 +1,5 @@
 const Postcode = require('postcode');
+const messages = require('../lib/messages');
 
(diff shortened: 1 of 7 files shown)


st3v3nhunt:
Closing as this has been superseded by #117

PR #118, opened 2016-10-17
:metal: Search from point with greater choice

st3v3nhunt:
3 orgs are returned in the middle of the night e.g. 2-4 am from this point. The other point was too far away from a good selection of orgs with wide ranging opening times.

And, I noticed 2 instances of moment were being used to check for the opening times. This could have caused it to have been open for the check in one instant and then shut in the other. Now the same value is used for both checks.

I've also included the change to use the latest version of the moment lib repo since the changes are all merged into master and are working.

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,7 +41,7 @@
     "geolib": "^2.0.21",
     "memory-cache": "^0.1.6",
     "moment": "^2.14.1",
-    "moment-opening-times": "git://github.com/nhsuk/moment-opening-times.git#feature/change-opening-times-format",
+    "moment-opening-times": "git://github.com/nhsuk/moment-opening-times.git",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -41,7 +41,7 @@
     "geolib": "^2.0.21",
     "memory-cache": "^0.1.6",
     "moment": "^2.14.1",
-    "moment-opening-times": "git://github.com/nhsuk/moment-opening-times.git#feature/change-opening-times-format",
+    "moment-opening-times": "git://github.com/nhsuk/moment-opening-times.git",
```
neilbmclaughlin: I've tagged the release in github so you can specify a version (e.g. git://github.com/nhsuk/moment-opening-times.git#1.0.0)

PR #124, opened 2016-10-18
:arrow_backward: Revert to moment lib pre-parser

st3v3nhunt:


st3v3nhunt:
Moments lib is now updated to the not have the parser. This is no longer required.

PR #130, opened 2016-10-20
:pager: Add styled 404 page

### LINKED ISSUE #125: We don't have a specific 404 error page
At the moment we just have a general catch all error page.

Should handle 404 separately.

### LINKED ISSUE #126: App does not handle postcode.io unavailability gracefully.
In simulating what would happen if postcode.io was unavailable (by disconnecting wifi) the behaviour of the post code search was not as expected.

It returns a 'Postcode Error' message ('Use a valid postcode') on the search page.

neilbmclaughlin:
Also added code tests for postcode to location lookup fails

Fixes #126
Fixes #125

### DIFF
--- a/app/views/error-404.nunjucks
+++ b/app/views/error-404.nunjucks
@@ -7,7 +7,7 @@ Page not found
 {% block content %}
   <div class="reading-width">
 		<p>If you have entered a web address check it was correct. You can <a href="//nhs.uk/">browse from the NHS.UK home page</a>.</p>
-		<p><a href="javscript:history.go(-1)" class="link-back">Back</a></p>
+		<p><a href="javascript:history.go(-1)" class="link-back">Back</a></p>
   </div>
 </div>
 {% endblock %}
--- a/app/views/error-404.nunjucks
+++ b/app/views/error-404.nunjucks
@@ -1,0 +1,13 @@
+{% extends 'layout.nunjucks' %}
+
+{% block header %}
+Page not found  
+{% endblock %}
+
+{% block content %}
+  <div class="reading-width">
+		<p>If you have entered a web address check it was correct. You can <a href="//nhs.uk/">browse from the NHS.UK home page</a>.</p>
+		<p><a href="javscript:history.go(-1)" class="link-back">Back</a></p>
+  </div>
+</div>
+{% endblock %}
--- a/app/views/error.nunjucks
+++ b/app/views/error.nunjucks
@@ -3,14 +3,3 @@
 {% block header %}
 A server error has occured.
 {% endblock %}
-
-{% block content %}
-<p>This error page should only appear in non-production envirnoments.</p>
-<p>In production Akamai will serve up it's own error pages for 500 errors.</p>
-<div id="error-message">
-  <div>{{ message }}</div>
-  <div>{{ error.status }}</div>
-  <pre>{{ error.stack }}</pre>
-</div>
-</div>
-{% endblock %}
--- a/app/views/error.nunjucks
+++ b/app/views/error.nunjucks
@@ -1,13 +1,13 @@
 {% extends 'layout.nunjucks' %}
 
 {% block header %}
-Ooops...an error has occured.
+A server error has occured.
 {% endblock %}
 
 {% block content %}
-<div onclick="(function myFunction() { var error = document.getElementById('error-message'); error.style.display = error.style.display == 'none' ? 'block' : 'none'; })()">
-Looks like one of the techies has pulled a sickie. Please try again or get in touch if it keeps happening.
-<div id="error-message" style="display:none;">
+<p>This error page should only appear in non-production envirnoments.</p>
+<p>In production Akamai will serve up it's own error pages for 500 errors.</p>
+<div id="error-message">
   <div>{{ message }}</div>
   <div>{{ error.status }}</div>
   <pre>{{ error.stack }}</pre>
--- a/config/express.js
+++ b/config/express.js
@@ -37,10 +37,9 @@ module.exports = (app, config) => {
   app.use(constants.SITE_ROOT, express.static(`${config.root}/public`));
 
   app.use(constants.SITE_ROOT, router);
-  app.use(constants.SITE_ROOT, (req, res, next) => {
-    const err = new Error('Not Found');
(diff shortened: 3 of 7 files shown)


neilbmclaughlin:
Code rebased. Some conflicts resolved around the integration tests for redirection which would be worth looking at.

app/views/error-404.nunjucks
```diff
@@ -0,0 +1,13 @@
+{% extends 'layout.nunjucks' %}
+
+{% block header %}
+Page not found  
+{% endblock %}
+
+{% block content %}
+  <div class="reading-width">
+		<p>If you have entered a web address check it was correct. You can <a href="//nhs.uk/">browse from the NHS.UK home page</a>.</p>
+		<p><a href="javscript:history.go(-1)" class="link-back">Back</a></p>
```
st3v3nhunt: Typo, missing `a`. Should be `javascript:history.go(-1)`

test/integration/app.js
```diff
@@ -151,42 +144,85 @@ describe('The results page', () => {
...
+              });
+          });
 
-    const postcodesioScope =
-    nock('https://api.postcodes.io')
-      .get(`/outcodes/${invalidPostcodePassingRegex}`)
-      .times(1)
-      .reply(404, notFoundResponse);
+        it('should validate the postcode and return an error message', (done) => {
+          nock('https://api.postcodes.io')
```
st3v3nhunt: There should be no requests going to postcodes.io as the validation will be done by the postcode regex checker. The `nock` can be removed.

test/integration/app.js
```diff
@@ -151,42 +144,85 @@ describe('The results page', () => {
...
-        .query({ location: invalidPostcode })
-        .end((err, res) => {
-          checkHtmlResponse(err, res);
-          expect(res.text).to.contain(errorMessage);
-          done();
-        });
+      it('postcode server error should return an error', (done) => {
+        const postcodesioScope =
+          nock('https://api.postcodes.io')
+          .get('/postcodes/AB123CD')
```
st3v3nhunt: Can the postcode `AB123CD` be replaced by the variable to make it clear the nock is being used for the request on line 212?

config/express.js
```diff
@@ -37,10 +37,9 @@ module.exports = (app, config) => {
   app.use(constants.SITE_ROOT, express.static(`${config.root}/public`));
 
   app.use(constants.SITE_ROOT, router);
-  app.use(constants.SITE_ROOT, (req, res, next) => {
-    const err = new Error('Not Found');
-    err.statusCode = 404;
-    next(err);
+  app.use(constants.SITE_ROOT, (req, res) => {
+    res.status(404);
+    res.render('error-404', {});
```
st3v3nhunt: Does passing the empty object serve any purpose?

test/unit/lib/postcodes.js
```diff
@@ -77,6 +77,40 @@ describe('Postcodes', () => {
       });
     });
 
+    describe('server errors', () => {
+      it('should return and error when postcode service throws a 500 error', (done) => {
```
st3v3nhunt: Typo? `and` should be `an`?

test/unit/lib/postcodes.js
```diff
@@ -77,6 +77,40 @@ describe('Postcodes', () => {
...
+        const postcode = 'AB123CD';
+        const postcodeRes = { locals: { location: postcode } };
+        const errorMessage = 'getaddrinfo ENOTFOUND api.postcodes.io api.postcodes.io:443';
+        const scope =
+          nock('https://api.postcodes.io')
+          .get(`/postcodes/${postcode}`)
+          .replyWithError({ message: errorMessage });
+
+        postcodes.lookup(postcodeRes, (err) => {
+          expect(err.message).to.be.equal(errorMessage);
```
st3v3nhunt: Is it worth adding a check for the type of error too?

test/unit/lib/postcodes.js
```diff
@@ -77,6 +77,40 @@ describe('Postcodes', () => {
...
+      it('should return and error when postcode service throws a 500 error', (done) => {
+        const postcode = 'AB123CD';
+        const postcodeRes = { locals: { location: postcode } };
+        const scope =
+          nock('https://api.postcodes.io')
+          .get(`/postcodes/${postcode}`)
+          .reply(500);
+
+        postcodes.lookup(postcodeRes, (err) => {
+          expect(err.message).to.be.equal('Postcode service error: 500');
```
st3v3nhunt: Is it worth adding a check for the type of error too?

app/views/error.nunjucks
```diff
@@ -1,13 +1,13 @@
...
 {% block header %}
-Ooops...an error has occured.
+A server error has occured.
 {% endblock %}
 
 {% block content %}
-<div onclick="(function myFunction() { var error = document.getElementById('error-message'); error.style.display = error.style.display == 'none' ? 'block' : 'none'; })()">
-Looks like one of the techies has pulled a sickie. Please try again or get in touch if it keeps happening.
-<div id="error-message" style="display:none;">
+<p>This error page should only appear in non-production envirnoments.</p>
```
neilbmclaughlin: The aim here was to make this page substantially different from the Akamai one so, for example, if the Akamai config was incorrect in a particular environment it would be immediately obvious. I didn't want this page to leak through into production given that it displays error message.

I guess the answer is to remove the display of error messages so we have to rely on logging then if it does get displayed then it is not a problem.

neilbmclaughlin: Thoughts?

config/express.js
neilbmclaughlin: Nope, removed.

app/views/error.nunjucks
st3v3nhunt: Removing the error message would be a good idea.

I understand the desire to be able to immediately see if there is a problem with the config using this page but I think we should probably be using other things to test that. If you really wanted to do that, perhaps just have something in the source of the page e.g. `generated by app` as the differentiator?

I'll hold off merging until the page has been stripped back.

st3v3nhunt:
Just one comment on the error page.

neilbmclaughlin:
Exception details removed from error page and ready for merge.

PR #132, opened 2016-10-20
:bug: Fix broken test

neilbmclaughlin:
Just a illustration of the fix needed not intended for merge.
Will need tests.


st3v3nhunt:
You could argue there is already a test given that is how this change has been identified!

However, the bigger issue here is the fact this bug should have been highlighted much sooner and there are tests that would have caught this right from the off (the test that ended up failing now). The reason this hasn't happened is because `openServices` and `nearbyServices` are the same objects! My Bad. I'll get it sorted.

st3v3nhunt:
Closing as #136 sorts it out

PR #136, opened 2016-10-21
:black_large_square: Ensure openServices are not refs to nearbyServices

### LINKED ISSUE #131: Broken travis build for PR128
I'm guessing it's a time issue again but no time to investigate (pun intended!).

### LINKED ISSUE #133: Clone objects rather than use copies of them

st3v3nhunt:
Fixes #133 
Fixes #131

### DIFF
--- a/test/unit/lib/utils.js
+++ b/test/unit/lib/utils.js
@@ -1,0 +1,29 @@
+const chai = require('chai');
+const utils = require('../../../app/lib/utils');
+
+const expect = chai.expect;
+
+describe('utils', () => {
+  describe('deepClone', () => {
+    it('should clone value properties', () => {
+      const objToClone = { a: 'a' };
+
+      const clone = utils.deepClone(objToClone);
+
+      expect(clone).to.deep.equal(objToClone);
+
+      clone.a = 'new value';
+      expect(clone).to.not.deep.equal(objToClone);
+    });
+    it('should clone object properties', () => {
+      const objToClone = { a: { b: 'b' } };
+
+      const clone = utils.deepClone(objToClone);
+
+      expect(clone).to.deep.equal(objToClone);
+
+      clone.a = 'new value';
+      expect(clone).to.not.deep.equal(objToClone);
+    });
+  });
+});
--- a/app/lib/getPharmacies.js
+++ b/app/lib/getPharmacies.js
@@ -2,6 +2,7 @@ const OpeningTimes = require('moment-opening-times');
 const moment = require('moment');
 const geolib = require('geolib');
 const assert = require('assert');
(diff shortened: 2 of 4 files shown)


neilbmclaughlin:
Did you consider writing a failing test which reproduced the problem before implementing the fix?

test/unit/lib/utils.js
```diff
@@ -0,0 +1,29 @@
+const chai = require('chai');
+const utils = require('../../../app/lib/utils');
+
+const expect = chai.expect;
+
+describe('utils', () => {
+  describe('deepClone', () => {
```
neilbmclaughlin: Since `JSON.parse(JSON.stringify(obj));` is such a common implementation of a 'poor mans clone' isn't this just testing javascript implementation.

st3v3nhunt:
The test was failing already and when the implementation is removed it continues to fail so no new test.

test/unit/lib/utils.js
st3v3nhunt: Maybe, but the implementation can change and we have tests that cover the required functionality.

Why the comment about `poor man's clone`?

neilbmclaughlin:
My question about having a failing test was related to my assumption that if the tests were running OK on master and the branch before the merge then the most likely cause of a failing test was down to a timing issue. I assumed that the test failed when it had previously been passing was because it was being run in the evening. Do you not think this was the case? I've tried to replicate the failing test by checking out commit #341c9f1 and have not been able to but that is probably me just being daft.

test/unit/lib/utils.js
neilbmclaughlin: Nothing derogatory implied by poor man just that it is a partial object clone (in that it just does public properties not methods) which is easy to implement.

st3v3nhunt: I didn't take it as derogatory, merely wondering if there was some additional context I didn't understand. I was wondering if there was a better (rich man's) version...

I was going to use `Object.assign({}, obj)` but didn't because it is a shallow clone. And yes, `JSON.parse(JSON.stringify(obj))` isn't very clever and doesn't include functions but it only needs to deal with the data so I figured it would be fine. Hence some very basic tests to check it does the data clone and the implementation could be swapped out if/when required.

st3v3nhunt:
Yeah, the time caused the issue but the issue is nothing to do with the time. (as you identified) all the time did was reduce the commonality between the open and the nearby services to such a degree that none of the nearby ones were open.

The actual issue is due to the open objects being only a reference to the nearby ones (or vice versa) to the same objects. So, I made a change to clone the open services rather than have a copy of them and the tests for the open results failed. That was my failing test which lead me to get it fixed.

PR #140, opened 2016-10-21
:1234: Changed results html to ordered list

mcheung-nhs:


st3v3nhunt:
I don't know if you want to update the version of the assets in the view? In theory it shouldn't matter as Akamai is handling the cache expiration.
I'd quite like to check if that does work on this and then we'll know for sure for next time. What do you reckon?

PR #143, opened 2016-10-24
:eyes: Added aria labels and associated ids for errors

mcheung-nhs:


mcheung-nhs:
Added aria labels to the enter postcode form field when there is an error so screen readers are told what the problem is.

Moved the internal "Use a valid postcode" link at the top of the content to point to the container div of the input field rather than to bring focus to the input field itself so as to work better with VoiceOver. When directly focusing in to the form field, there is a loss of focus and VoiceOver doesn't know to bring up the keyboard so the user can entering something.

neilbmclaughlin:
Just listened to this with ChromeVox and I can see the change is there and it seems to report the postcode errors. Just the link text rather than the descriptive text is read which is perhaps wrong. For example 'A valid postcode is required to progress' when no post code is entered vs 'fgdhgdfg is not a valid postcode, please try again' when an invalid postcode is entered.

However I don't seem to be able to navigate effectively as I don't know the navigation shortcuts so don't feel able to assess whether this fix is complete. Perhaps if you could spend some time showing me the change and we can talk it through.

mcheung-nhs:
Updated changes made in branch fix/accessibility

PR #147, opened 2016-10-25
:sound: Ensure error messages are consistent

neilbmclaughlin:
Also changed page title to ensure context is given to screen readers

### DIFF
--- a/app/views/error-404.nunjucks
+++ b/app/views/error-404.nunjucks
@@ -10,7 +10,7 @@
 
 {% block content %}
   <div class="reading-width">
-    <p>If you have entered a web address check it was correct. You can <a href="//nhs.uk/">browse from the NHS.UK home page</a>.</p>
+    <p>If you have entered a web address check it was correct. You can <a href="http://nhs.uk/">browse from the NHS.UK home page</a>.</p>
     <p><a href="javascript:history.go(-1)" class="link-back">Back</a></p>
   </div>
 {% endblock %}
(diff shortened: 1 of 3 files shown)


neilbmclaughlin:
Just spotted that the html contains tabs - will correct and force push

neilbmclaughlin:
Done and ready for review.

@mcheung-nhs @ermlikeyeah wondering if the `<div class="reading-width">` should be in layout.nunjucks since it is replicated in both error pages although find-help.nunjucks doesn't use it. Any thoughts?

st3v3nhunt:
@neilbmclaughlin Some conflicts need sorting out please

app/views/error-404.nunjucks
```diff
@@ -1,13 +1,16 @@
...
 {% block header %}
-Page not found  
+  {{ message }}
 {% endblock %}
 
 {% block content %}
   <div class="reading-width">
-		<p>If you have entered a web address check it was correct. You can <a href="//nhs.uk/">browse from the NHS.UK home page</a>.</p>
-		<p><a href="javascript:history.go(-1)" class="link-back">Back</a></p>
+    <p>If you have entered a web address check it was correct. You can <a href="//nhs.uk/">browse from the NHS.UK home page</a>.</p>
```
st3v3nhunt: The href here needs to be `http://www.nhs.uk` given the current site doesn't work over `https`

PR #148, opened 2016-10-25
:closed_lock_with_key: Set Azure to redirect all traffic to https

### LINKED ISSUE #146: Redirect http traffic to https
Ensure all traffic if redirected to `https`

st3v3nhunt:
Setting up environment in Azure to test the change before merging where it will be automatically deployed to the staging slot.

New slot added and pointing to this branch i.e. `feature/https-all-traffic`
http://connecting-to-services-steve-test.azurewebsites.net/

The site automatically redirects to https://connecting-to-services-steve-test.azurewebsites.net/finders/find-help

Fixes #146

### DIFF
--- a/web.config
+++ b/web.config
@@ -1,0 +1,61 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!--
+     This configuration file is required if iisnode is used to run node processes behind
+     IIS or IIS Express.  For more information, visit:
+
+     https://github.com/tjanczuk/iisnode/blob/master/src/samples/configuration/web.config
+-->
+
+<configuration>
+  <system.webServer>
+    <!-- Visit http://blogs.msdn.com/b/windowsazure/archive/2013/11/14/introduction-to-websockets-on-windows-azure-web-sites.aspx for more information on WebSocket support -->
+    <webSocket enabled="false" />
+    <handlers>
+      <!-- Indicates that the server.js file is a node.js site to be handled by the iisnode module -->
+      <add name="iisnode" path="app.js" verb="*" modules="iisnode"/>
+    </handlers>
+    <rewrite>
+      <rules>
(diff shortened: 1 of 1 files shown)


web.config
```diff
@@ -0,0 +1,71 @@
...
+
+     https://github.com/tjanczuk/iisnode/blob/master/src/samples/configuration/web.config
+-->
+
+<configuration>
+  <system.webServer>
+    <!-- Visit http://blogs.msdn.com/b/windowsazure/archive/2013/11/14/introduction-to-websockets-on-windows-azure-web-sites.aspx for more information on WebSocket support -->
+    <webSocket enabled="false" />
+    <handlers>
+      <!-- Indicates that the server.js file is a node.js site to be handled by the iisnode module -->
```
neilbmclaughlin: Comment is incorrect. Correct or, preferably, delete.

st3v3nhunt:
The file is the autogenerated one with the addition of the rule. I'll make it clearer.

st3v3nhunt:
Do you want it with comments or without?

neilbmclaughlin:
Ah, I did wonder about if it might be :)

Not worth doing it now but in a similar situation would be worth adding as one commit and updating as another.

st3v3nhunt:
It is easy enough to do and probably a good idea

neilbmclaughlin:
Regarding comments - I don't think this comment adds anything.

st3v3nhunt:
Do you want a fresh auto generated version or one without that comment? Or to remove that comment later?

neilbmclaughlin:
Assuming the autogenerated one already contains the erroneous comment I'd commit exactly as it is with the error. The 2nd commit for the http redirection could then include the removal of the comment with a commit note.

st3v3nhunt:
Fine grained changes detailed through commit history as requested

PR #150, opened 2016-10-26
:sleeping: Use debug rather than console logging

st3v3nhunt:
In order to control the output of the logging/debugging information I've introduced the `debug` package as a replacement for using `console`. This allows specific information based on the namespace of debug to be turned via the environment variable `DEBUG`.

Additional information is available on https://www.npmjs.com/package/debug

I've the suggested convention of name-spacing the application and the various modules within it e.g. `finders:thing`


st3v3nhunt:
@neilbmclaughlin Fair point on the start/end. I've added it where applicable. I'd left it off before because there is no ordering/association and if a new debug was added in between a start and end the time becomes incorrect. At least as a single number where no addition of previous times would be required...

st3v3nhunt:
[Improve application logging](https://trello.com/c/cTtUddvi/451-improve-application-logging)

PR #151, opened 2016-10-26
:relieved: Clone open obj once opening info has been added

st3v3nhunt:
This fixes a bug that was introduced when [open objects started to be cloned](https://github.com/nhsuk/connecting-to-services/commit/a0575433f36e783f476e32e168602825385f8980). The problem being that both `isOpen` and `openingTimesMessage` were not added to the object being cloned before it was cloned. This can be seen on the website by simply visiting the results page for the first time with a new postcode. No opening time message will be displayed. This changes fixes that so the message should be displayed immediately and consistently (as it has been in the past).

### DIFF
--- a/app/lib/getPharmacies.js
+++ b/app/lib/getPharmacies.js
@@ -61,21 +61,23 @@ function nearby(searchPoint, geo, limit) {
 
     if (openingTimes) {
       const openingTimesMoment = new OpeningTimes(item.openingTimes.general, 'Europe/London');
+
       openingTimesMessage = openingTimesMoment.getOpeningHoursMessage(now);
       isOpen = openingTimesMoment.isOpen(now);
-      if (isOpen && openServiceCount < numberOfOpenToReturn) {
-        openServiceCount++;
-        openServices.push(utils.deepClone(item));
-      }
     } else {
       openingTimesMessage = 'Call for opening times';
       isOpen = false;
     }
-    serviceCount++;
-    /* eslint-disable no-param-reassign */
+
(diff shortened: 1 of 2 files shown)


app/lib/getPharmacies.js
```diff
@@ -61,21 +61,23 @@ function nearby(searchPoint, geo, limit) {
...
-    serviceCount++;
-    /* eslint-disable no-param-reassign */
+
     item.openingTimesMessage = openingTimesMessage;
     item.isOpen = isOpen;
-    /* eslint-enable no-param-reassign */
+
+    if (isOpen && openServiceCount < numberOfOpenToReturn) {
+      openServiceCount++;
+      openServices.push(utils.deepClone(item));
```
neilbmclaughlin: Just an observation for future reference (no action required) ...

our current implementation of deepClone would preclude us from using the openingTimes object on the service (e.g. `item.openingTimes = new OpeningTimes(...)`)

test/unit/lib/getPharmacies.js
```diff
@@ -74,7 +74,24 @@ describe('Nearby', () => {
     });
 
     it('should return the opening times message and open state', () => {
-      const results = pharmacies.nearby(searchPoint, geo).openServices;
+      const alwaysOpenOrg = {
```
neilbmclaughlin: I'm guessing we don't have a closed test due to the fact we can't currently write a reliable test which will not fail depending on the time the test is run. I've added an issue for this (#154)

PR #153, opened 2016-10-27
:globe_with_meridians: Changed map link url

mcheung-nhs:


st3v3nhunt:
@mcheung-nhs Are you able to test this on an Android device? I've checked on desktop and an iOS device and it is working as I would expect with Google Maps being used on the iOS device and the website on the desktop

mcheung-nhs:
Confirmed to be working on a Moto E, Nexus 5 and Samsung Galaxy S4 devices.
Checked on Browserstack with Samsung Galaxy S7 and S6 and S5 and works for these too.

PR #155, opened 2016-10-27
:eyes: Aria labels and error message fixes

mcheung-nhs:


mcheung-nhs:
1. Added aria labels to the enter postcode form field so when there is an error screen readers are told there is a problem.

2. Moved the internal "Use a valid postcode" link at the top of the content to point to the error message rather than to bring focus to the input field itself so as to work better with VoiceOver. When directly focusing in to the form field, there is a loss of focus and VoiceOver doesn't know to bring up the keyboard so the user can entering something.

3. Removed autofocus on the enter postcode input field because android TalkBack ignored any content above it, which may have some context and be relevant.

PR #158, opened 2016-10-28
:football: Use Bunyan for logging

st3v3nhunt:
I've removed both `debug` and `morgan` and am now solely using `bunyan` for logging. This means the logs are output all of the time rather than with `debug` where it had to be turned on. To provide similar functionality the levels of the logging can be used. I've set the default level to be `bunyan.INFO` and provided a script to output `bunyan.DEBUG` when running locally i.e. `npm run dev-watch-debug`.

Using `bunyan` paves the way to push the logs into Splunk.


st3v3nhunt:
[Improve application logging](https://trello.com/c/cTtUddvi/451-improve-application-logging)

app/lib/postcodes.js
```diff
@@ -36,11 +36,11 @@ function lookup(res, next) {
           next();
           break;
         case 404:
-          debugPostcodes({ url, response: postcodeRes.statusCode });
+          log.warn({ res: postcodeRes }, '404 from postcodes.io');
```
neilbmclaughlin: The failing postcode can be seen in the subsequent debug line of the log (http://localhost:3000/finders/results?location=*HG5+0JK*&open=true&context=) but it might be easier when scanning the log file to have both the URL and status code in the message.

PR #159, opened 2016-10-31
Feature/bunyan splunk logger

st3v3nhunt:
When the application is running in `production` mode the logs should be sent to Splunk using the npm package [`splunk-bunyan-logger`](https://www.npmjs.com/package/splunk-bunyan-logger).

Until such a time when the url and token is known (in order to setup and test the configuration) this change should not be merged as it will break the production environment!

I'll extract the logger specific utils out into their own `logger-utils` file at some point too.


st3v3nhunt:
[Make application logs available via Splunk](https://trello.com/c/oKF3n5Qh/548-make-application-logs-available-via-splunk)

st3v3nhunt:
Updating `request` from `2.60.0` to `2.78.0` to get rid of possible remote memory exposure https://snyk.io/vuln/npm:request:20160119

st3v3nhunt:
Ready for review and in theory would be ok to be merged.

I've added `SPLUNK_HEC_TOKEN` and `SPLUNK_HEC_URL` to both staging and prod azure environments along with Heroku so logs should start to come through once it has been deployed into those environments.

There is still work to do but that can be done after this has been reviewed:

* Integrate snyk into CI / commit work flow
* Increase the logging output e.g. page loads (including static assets)

PR #165, opened 2016-11-02
:rocket: Add simple performance test

st3v3nhunt:
This is a very simple test that will currently only work when run on a dev machine due to the value of `target` in the test script. Target can be overridden on the [command line](https://artillery.io/docs/cli_reference.html) so it could be changed based on the environment it need to run against. Or, different environments can be specified in the script and chosen at runtime. More information available at https://artillery.io/docs/index.html

Things to think about are how this would be work in practice as part of the CI pipeline. Atm Heroku review apps are used but they are deployed by Heroku. That information (the URL) is not available in Travis (but could be calculated). However, the deployments could be moved to become Travis's responsibility, but, I don't think review apps are able to be used taking that approach.

It might be the case that the tests are run against persistent environments and the deployment pipeline is changed e.g. use `pre-release` branch for all working feature merges and only once everything is working on that branch does it get merged into `master` (which is still used as the main site's deployment source).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 node_modules/
 public/components
 coverage/
+reports/
 
 .sass-cache
 npm-debug.log
@@ -10,3 +11,6 @@ npm-debug.log.*
 .coveralls.yml
 
 *.swp
+
+# Use .keep files to keep empty folders
+!.keep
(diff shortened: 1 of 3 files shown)


st3v3nhunt:
I'm going to close this PR as I've found an easier to way to generate the information I wanted to get out of it. I didn't really want to have performance tests running what I really wanted was to be able to get an idea of whether changes made had made any discernible difference to the execution time of the existing tests.

In order to do that I have been running the tests with the xunit reporter, saving a copy of the report and comparing it between runs:` npm run test -- --reporter xunit --grep <filter-word> > <file-name>.xml`

Issue #169, opened 2016-11-03
Remove hard coding of font CDN URL


neilbmclaughlin:
The URL is hard coded in express config and in the css files.
Should be in an env var.

st3v3nhunt:
Fixed by #271

PR #167, opened 2016-11-03
:closed_lock_with_key: Refine security headers

neilbmclaughlin:

PR #170, opened 2016-11-04
Feature/more headers

st3v3nhunt:
This fixes a couple of error messages seen in the browser console.

|Page|Error|Fix|
|---------|---------|-------|
|https://connecting-to-services-pr-167.herokuapp.com/finders/find-help|Refused to execute inline script because it violates the following Content Security Policy directive: "script-src 'self' data: www.google-analytics.com s.webtrends.com statse.webtrendslive.com static.hotjar.com script.hotjar.com cdn.jsdelivr.net".|https://github.com/nhsuk/connecting-to-services/pull/170/files#diff-1c3fafa210dc8117aaa1accc8b43316bR31|
|https://connecting-to-services-pr-167.herokuapp.com/finders/find-help|Refused to apply inline style because it violates the following Content Security Policy directive: "style-src 'self' fast.fonts.net https://dhrlmnmyf2njb.cloudfront.net/".|https://github.com/nhsuk/connecting-to-services/pull/170/files#diff-1c3fafa210dc8117aaa1accc8b43316bR49|
|https://connecting-to-services-pr-167.herokuapp.com/finders/results?location=ls1&open=true&context=|Refused to execute inline script because it violates the following Content Security Policy directive: "script-src 'self' data: www.google-analytics.com s.webtrends.com statse.webtrendslive.com static.hotjar.com script.hotjar.com cdn.jsdelivr.net".|https://github.com/nhsuk/connecting-to-services/pull/170/files#diff-1c3fafa210dc8117aaa1accc8b43316bR31|

### DIFF
--- a/config/express.js
+++ b/config/express.js
@@ -28,6 +28,7 @@ module.exports = (app, config) => {
       ],
       scriptSrc: [
         '\'self\'',
+        '\'unsafe-inline\'',
         'data:',
         'www.google-analytics.com',
         's.webtrends.com',
@@ -45,6 +46,7 @@ module.exports = (app, config) => {
       ],
       styleSrc: [
         '\'self\'',
+        '\'unsafe-inline\'',
         'fast.fonts.net',
         'https://dhrlmnmyf2njb.cloudfront.net/',
       ],
--- a/config/express.js
+++ b/config/express.js
@@ -46,9 +46,11 @@ module.exports = (app, config) => {
       styleSrc: [
         '\'self\'',
         'fast.fonts.net',
+        'dhrlmnmyf2njb.cloudfront.net',
       ],
       fontSrc: [
         'fast.fonts.net',
+        'https://dhrlmnmyf2njb.cloudfront.net/',
       ],
       connectSrc: [
         '\'self\'',
--- a/config/express.js
+++ b/config/express.js
@@ -46,7 +46,7 @@ module.exports = (app, config) => {
       styleSrc: [
         '\'self\'',
         'fast.fonts.net',
-        'dhrlmnmyf2njb.cloudfront.net',
+        'https://dhrlmnmyf2njb.cloudfront.net/',
       ],
       fontSrc: [
         'fast.fonts.net',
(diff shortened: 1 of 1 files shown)


neilbmclaughlin:
Nice spot on the breaking change. I removed the inline script directive as it seemed risky - AFAIK that would be the mechanism for XSS. Do you think it would be better for the inline js to be in files rather than adding this directive?

st3v3nhunt:
Yes, moving away from having any inline script is going to be a much better solution for countering XSS issues. And in fact is necessary for [Chrome v45 and previous versions](https://developer.chrome.com/extensions/contentSecurityPolicy#relaxing).
It will also force a cleaner separation between form and function.

What do you reckon - get that done rather than merging this change?

neilbmclaughlin:
I think it shouldn't take long so I'd vote for moving them into files.

st3v3nhunt:
Yup, ok. I'll get that started

st3v3nhunt:
Thinking about the details of what is required here is a bit more involved than I first thought. The inline script contain dynamic values that would require files to be generated on app start up and have their file name hashed to avoid issues with caching. I'll raise an issue to get that sorted at some point.

st3v3nhunt:
Thinking about the details of what is required here is a bit more involved than I first thought. The inline script contain dynamic values that would require files to be generated on app start up and have their file name hashed to avoid issues with caching. I'll raise an issue to get that sorted at some point.

neilbmclaughlin:
Commit from this branch now in PR #167.

st3v3nhunt:
[Add http security headers](https://trello.com/c/6aJCdJP6/561-add-http-security-headers)

config/express.js
```diff
@@ -45,10 +46,13 @@ module.exports = (app, config) => {
       ],
       styleSrc: [
         '\'self\'',
+        '\'unsafe-inline\'',
         'fast.fonts.net',
+        'https://dhrlmnmyf2njb.cloudfront.net/',
```
teneightfive: @st3v3nhunt just spotted this after looking at #167 

With this URL it would be better to store it in an environment variable as this could well change per environment or at any time if we change where the font/assets are served from

st3v3nhunt: @neilbmclaughlin got an issue raised for it #169

teneightfive: cool 👍

PR #174, opened 2016-11-04
Feature/bunyan splunk

st3v3nhunt:
Same as #159 

Seeing if this will prompt a new security scan by snyk since a package update


st3v3nhunt:
Nope, it has nothing to do with the package that has been updated. It is a dependency that has caused it! 👊

PR #179, opened 2016-11-07
:ferris_wheel: Mute dep that has been patched by synk

st3v3nhunt:
In `splunk-bunyan-logger` there is a vulnerable dependency. This has been patched by Synk but bitHound doesn't know that.
This change mutes that dependency warning (as it has been dealt with).

### DIFF
--- a/.bithoundrc
+++ b/.bithoundrc
@@ -53,6 +53,9 @@
   "dependencies": {
     "unused-ignores": [
       "snyk"
+    ],
+    "mute": [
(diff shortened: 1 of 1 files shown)


neilbmclaughlin:
How do we ensure that the muting is removed when the package has been upgraded.

If we don't do we run the risk of missing a future insecurity.

I guess if snyk becomes our tool for detecting insecurities then this doesn't matter.

We are back to the issue of what does bithound give us again?

PR #187, opened 2016-11-08
:bullettrain_front: Use optimised OpeningTimes

neilbmclaughlin:
PR https://github.com/nhsuk/moment-opening-times/pull/39 will need merging and packaging before this is merged.
The moment-opening-times dependency in packages.json will then also need updating.

**Timings in connecting-to-services with the optimised OpeningTimes code:**

experimental/exceptional-opening-times
  Nearby
    happy path
      ✓ should return an object with nearby and open services (132ms)

master
  Nearby
    happy path
      ✓ should return an object with nearby and open services (101ms)

### DIFF
--- a/app/lib/getOpeningTimesMessage.js
+++ b/app/lib/getOpeningTimesMessage.js
@@ -3,7 +3,7 @@ function closesAtMidnight(moment) {
   return (time === '00:00' || time === '23:59');
 }
 
-function getOpeningHoursMessage(status) {
+function getDayDescriptor(moment, referenceMoment) {
   const dayDescriptors = {
     sameDay: '[today]',
     nextDay: '[tomorrow]',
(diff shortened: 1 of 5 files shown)


neilbmclaughlin:
Note: getOpeningHoursMessage has been cut and pasted from OpeningTimes. It probably needs to be removed from the there and the tests need to be added to connecting-to-services.

neilbmclaughlin:
[Add exceptional opening times/dates](https://trello.com/c/8gHrmhJO/502-add-exceptional-opening-times-dates)

neilbmclaughlin:
I'm going to move the getOpeningHoursMessage method out into a separate lib file and add the tests.

neilbmclaughlin:
Open for review but not merge.

Once the moment-opening-times PR is merged and a release 2.0.0 created I'll make one last commit to update package.json to reference that release.

app/lib/getOpeningTimesMessage.js
```diff
@@ -0,0 +1,38 @@
+function getOpeningHoursMessage(status) {
+  if (status.isOpen) {
+    const closedNext = status.until;
```
st3v3nhunt: I've suggested `until` is renamed as I found it a bit confusing. I think something like `nextStateChange` would be more appropriate as it would mean `status.until` can be assigned to the variables as it is here and it wouldn't seem as odd.

app/lib/getOpeningTimesMessage.js
```diff
@@ -0,0 +1,38 @@
...
+    const closedDay = closedNext.calendar(status.moment, {
+      sameDay: '[today]',
+      nextDay: '[tomorrow]',
+      nextWeek: 'dddd',
+      lastDay: '[yesterday]',
+      lastWeek: '[last] dddd',
+      sameElse: 'DD/MM/YYYY',
+    });
+    return (
+      ((closedDay === 'tomorrow' && closedTime === '12:00 am')
```
st3v3nhunt: Could this be extracted out into a function to help explain what it is doing? Something like `isClosedTodayAndTomorrow()`

package.json
```diff
@@ -49,7 +49,7 @@
     "is-nan": "^1.2.1",
     "memory-cache": "^0.1.6",
     "moment": "^2.14.1",
-    "moment-opening-times": "git://github.com/nhsuk/moment-opening-times.git",
+    "moment-opening-times": "git://github.com/nhsuk/moment-opening-times.git#feature\/performance-improvements",
```
st3v3nhunt: I know this is mentioned in the PR notes but thought I'd highlight the fact it needs updating to whatever version is created when the lib changes are merged in so as not to forget to check.

app/lib/getOpeningTimesMessage.js
neilbmclaughlin: See comments in other PR. Happy to discuss the merits of different names via slack tomorrow.

neilbmclaughlin:
We should be good to go with this.

It would be worth getting Tom to re-run the performance tests once this is in staging. Even better if we were able to run them ourselves.

package.json
st3v3nhunt: This just needs changing to: `"moment-opening-times": "git://github.com/nhsuk/moment-opening-times.git#2.0.0",` and it is all good to merge.

PR #188, opened 2016-11-09
:o: Search for results in 10 mile circular segments

st3v3nhunt:
The rationale for this change is to speed up the search for organisations. By reducing the size of the initial area being searched and expanding it out until the required number of results has been found for most scenarios the time to get some results on screen is reduced. The changes I've made also mean the further distance a possible result can be is now 100 miles rather than the 50 it was searching previously.

I've made a change to the test data file so that the org found by searching for outcode `TR21` is always open.


st3v3nhunt:
[Speed up search](https://trello.com/c/VbwjwZb0/564-speed-up-search)

neilbmclaughlin:
This is a search results page from 00:09 tonight.

screen shot 2016-11-11 at 00 08 38

Assuming there is nothing wrong in principle with returning no results then there should probably be a message displayed. One for Joe/Nick.

st3v3nhunt:
@neilbmclaughlin I've added a card into the backlog to deal with getting this sorted as it is something happening already and nothing to do with the changes in the PR. Good spot.

test/unit/lib/getPharmacies.js
```diff
@@ -47,14 +48,48 @@ describe('Nearby', () => {
       });
     });
 
-    it('should get the number of results requested, ordered by distance', () => {
-      const requestedNumberOfResults = 5;
-      const results = pharmacies.nearby(searchPoint, geo, requestedNumberOfResults).nearbyServices;
+    it('should get the number of unique nearby open services requested, ordered by distance',
```
neilbmclaughlin: This seems to be testing 3 distinct things (number, uniqueness and distance).

It might be better to use an enclosing describe ('get open services') with 3 it() clauses.

There is probably also a need for a 4th it clause to expect on openness.

test/unit/lib/getPharmacies.js
```diff
@@ -47,14 +48,48 @@ describe('Nearby', () => {
...
+          expect(result.distanceInMiles).to.be.at.least(previousDistance);
+          previousDistance = result.distanceInMiles;
+        });
+      });
+
+    it('should get the number of unique nearby services requested, ordered by distance', () => {
+      const requestedNumberOfResults = 52;
+      const results =
+        pharmacies
+        .nearby(sillySearchPoint, geo, { nearby: requestedNumberOfResults })
```
neilbmclaughlin: I'm not clear about why this is a silly point and why it is being used here rather than search point.

neilbmclaughlin: Ah, now I know why. It's a 'Scilly' point. Still not sure why it is being used over search point. Suspect it is to do with remoteness. Maybe rename?

app/lib/utils.js
```diff
@@ -6,7 +6,13 @@ function deepClone(obj) {
   return JSON.parse(JSON.stringify(obj));
 }
 
+function removeDuplicates(myArr, prop) {
```
neilbmclaughlin: Nice, clean implementation.

Disappointing that javascript's Set data structure isn't usable for this due to lack of ability to use a custom equality  function.

app/lib/getPharmacies.js
```diff
@@ -86,18 +87,78 @@ function nearby(searchPoint, geo, limit) {
...
+
+  assert(geo, 'geo can not be null');
+  assert.equal(typeof (geo.nearBy), 'function', 'geo must contain a nearBy function');
+
+  const defaultNearbyLimit = 10;
+  const defaultOpenLimit = 3;
+
+  const limits = searchLimits || {};
+  limits.nearby = limits.nearby || defaultNearbyLimit;
+  limits.open = limits.open || defaultOpenLimit;
```
neilbmclaughlin: Consider asserting that limits are positive integers?

test/unit/lib/getPharmacies.js
```diff
@@ -13,18 +13,18 @@ const expect = chai.expect;
 describe('Nearby', () => {
   let geo = [];
   const searchPoint = { latitude: 53.797431921096, longitude: -1.55275457242333 };
-  const sillySearchPoint = { latitude: 49.9126167297363, longitude: -6.30890274047852 };
+  const remoteSearchPoint = { latitude: 49.9126167297363, longitude: -6.30890274047852 };
 
   describe('happy path', () => {
-    beforeEach('load data', () => {
+    before('load data', () => {
```
neilbmclaughlin: What is the rational for switching from `beforeEach` to `before`. I assume it's based on cache being read-only so beforeEach is unnecessary?

st3v3nhunt: Yeah, its gone as it wasn't doing anything useful. I *think* it was necessary at some point and when its been changed it just hasn't been updated.

st3v3nhunt:
@neilbmclaughlin I've updated the dependencies and the checks are all passing again. Ready for another review.

.eslintrc.json
```diff
@@ -15,6 +15,17 @@
       "import/no-extraneous-dependencies": [
         "error", { "devDependencies": true }
       ],
+      "no-plusplus": [
+        "error", { "allowForLoopAfterthoughts" : true }
```
neilbmclaughlin: I opted to allow the no-plusplus rule in moment-opening-times as it seemed better to not have a mix within the code.

What was your rational for going with allowForLoopAfterthoughts?

neilbmclaughlin: OK. In moment-opening-times there were just a couple of loops so the change was not a big deal and I therefore switched to +=/-= but it does mean that .eslintrc has diverged between the two repos. Just noting it here for future reference.

PR #195, opened 2016-11-24
:tea: Remove hard coded back link

st3v3nhunt:
Use the `referer` header value if present otherwise fall back to JS version. The reason for the fallback is due to the header not guaranteed to be sent.


st3v3nhunt:
[Make back link on find-help work for people who have not come from Stomach Ache condition](https://trello.com/c/6WMjMyAB/599-make-back-link-on-find-help-work-for-people-who-have-not-come-from-stomach-ache-condition)

PR #201, opened 2016-12-01
Feature/use api 2

st3v3nhunt:


app/middleware/getPharmacies.js
```diff
@@ -1,21 +1,37 @@
+const request = require('request');
 const log = require('../lib/logger');
-const cache = require('memory-cache');
-const pharmacies = require('../lib/getPharmacies');
 
 function getPharmacies(req, res, next) {
   const searchPoint = res.locals.coordinates;
-  const geo = cache.get('geo');
-  const limits = { nearby: 10, open: 3 };
+  // TODO: Implement this
```
st3v3nhunt: I'll get round to doing this as part of the changes to the results page

config/express.js
```diff
@@ -103,7 +103,7 @@ module.exports = (app, config) => {
   app.use(constants.SITE_ROOT, (err, req, res, next) => {
     const statusCode = err.statusCode || 500;
 
-    log.error({ err }, statusCode);
+    log.error(err, 'Error');
```
st3v3nhunt: I've changed this from the variable `statusCode` to the constant string `'Error'` for the reason that in the logs the message being the status code has no value as it is contained within the body of `err`. Additionally, if `err.statusCode` was not populated it would have been set to `500` which might not be accurate. Sending the status of `500` back from the application when an error has occurred and there is no known status seems to eb the correct thing to do so I've left it as is.

test/globalHooks.js
```diff
@@ -0,0 +1,30 @@
...
+
+  nock('https://api.postcodes.io')
+    .get(`/postcodes/${ls27ue}`)
+    .times(4)
+    .reply(200, ls27ueResponse);
+});
+
+after('Check all nocks have been called', () => {
+  expect(nock.pendingMocks().length).to.equal(0);
+  expect(nock.isDone()).to.equal(true);
```
neilbmclaughlin: I think this SO post [here](http://stackoverflow.com/questions/24553154/how-to-retest-same-url-using-mocha-and-nock) is describing a similar problem and an alternative solution. Is a possible solution to use `beforeEach(done)` and `afterEach(done)` to set up and clearAll the nocks? I only have a superficial understanding of the problem so may be talking rubbish!

app/middleware/getPharmacies.js
```diff
@@ -1,21 +1,34 @@
...
+  // const limits = { nearby: 10, open: 3 };
 
-  log.info('get-pharmacies-start');
-  const nearby = pharmacies.nearby(searchPoint, geo, limits);
-  log.info('get-pharmacies-end');
+  const latitude = searchPoint.latitude;
+  const longitude = searchPoint.longitude;
+
+  // TODO: Extract to ENV var
```
st3v3nhunt: Ha yeah, I thought I'd deleted that! In fact, it has been deleted in the latest version

README.md
```diff
@@ -55,6 +56,14 @@ environment.
 | `SPLUNK_HEC_ENDPOINT`            | [HTTP Event Collector endpoint](http://dev.splunk.com/view/event-collector/SP-CAAAE7H) |                          | In `production` |
 | `LOG_LEVEL`                      | [bunyan log level](https://github.com/trentm/node-bunyan#levels)                       | Depends on `NODE_ENV`    |                 |
```
st3v3nhunt: Good call, maybe in a separate commit given it hasn't changed?

test/globalHooks.js
st3v3nhunt: That post is about the same thing i.e nock is global. The solution of executing `nock.cleanAll()` is done here too and it might be possible to have the nock setup/teardown done back in the local tests. However, I prefer having it done globally to highlight the fact nock is global. I've also seen nocks still being active once the `cleanAll()` has been run so I have my doubts whether it would definitely work.

neilbmclaughlin: I saw the cleanAll but wasn't sure if you had investigated the usage of asynchronous beforeEach (i.e. with done). Not proposing a change just getting it clear in my head.

PR #203, opened 2016-12-05
Feature/front page changes

st3v3nhunt:
Some of the changes required for the finders page


st3v3nhunt:
[New Postcode Entry Page](https://trello.com/c/B0U3yIVt/610-new-postcode-entry-page)

app/lib/contexts.json
```diff
@@ -0,0 +1,7 @@
+{
+  "stomachAche": {
+    "context": "stomach-ache",
```
st3v3nhunt: Not sure if `context` is the best key name here? I was thinking `key` might be a better option but then thought that was too generic so left it as is.

app/lib/contexts.json
```diff
@@ -0,0 +1,7 @@
...
+    "url": "/symptoms/stomach-ache",
+    "text": "Back to information on stomach ache"
```
st3v3nhunt: `text` as a key isn't overly informative but i've not changed it because I didn't want to make it too specific until such a time as there are several contexts and there is a better understanding as to what this might become.

app/lib/contexts.json
```diff
@@ -0,0 +1,7 @@
+{
```
st3v3nhunt: The model for a `context` isn't great but as it is brand new and is going to change in the future I've left it as is until such a time it needs to change. Unless there is a pressing desire for those changes to be made sooner...

app/lib/contexts.json
neilbmclaughlin: I think I prefer 'key' but think we should make the change when we add next add a context

PR #208, opened 2016-12-09
:chart: Move to a single results page

st3v3nhunt:
As part of this commit I've away from having the nocks setup globally and brought them back to be closer to where they are used. The reason is because it is easier to have them done close to their usage, especially for figuring out when things are going correctly. And having the requests be really specific helps with that too.

As this change has started to use the API's ability to return the number of results requested those changes in the API will need to be available before this can be deployed. If this doesn't happen there will be the default values of 3 open and 10 nearby services returned.

**DONE:**

- [x] Remove duplication in results view for displaying the open and the nearby results
- [x] Remove any duplicate results due to having the same one in the open and nearby sets
- [x] Style the page
- [x] Add proper content for no results messaging


st3v3nhunt:
This https://github.com/nhsuk/nearby-services-api/pull/21 contains the changes to the API to allow limits to be specified by the client

app/middleware/getPharmacies.js
```diff
@@ -3,14 +3,9 @@ const log = require('../lib/logger');
...
   const searchPoint = res.locals.coordinates;
-  // TODO: Implement this
-  // const limits = { nearby: 10, open: 3 };
-
-  const latitude = searchPoint.latitude;
-  const longitude = searchPoint.longitude;
 
   const baseUrl = process.env.API_BASE_URL;
-  const url = `${baseUrl}/nearby?latitude=${latitude}&longitude=${longitude}`;
+  const url = `${baseUrl}/nearby?latitude=${searchPoint.latitude}&longitude=${searchPoint.longitude}&limits:results:open=1&limits:results:nearby=3`;
```
neilbmclaughlin: Useful to have a query parameter naming convention to group related parameters - why ':' rather than, say, '.' ?

Just checked and node doesn't just handle '.' (i.e. res.query.limits.results.open doesn't work need to use res.query['limits.results.open']). That would have been just too elegant :)

Should latitude and longitude have a similar convention (searchPoint:latitude)?

st3v3nhunt:
@neilbmclaughlin No longer WIP, this is good for some reviewing now

PR #210, opened 2016-12-13
:sparkles: Results page styling

mcheung-nhs:


app/views/results.nunjucks
```diff
@@ -9,34 +9,84 @@
...
+
+    {% else %}
+
+      <div class="results">
+
+        <div class="results-block">
+          <h2 class="results-none-open">There are no pharmacies open now within 20 miles of {{ location }}</h2>
+        </div>
+
+        <h2 class="">Next closest pharmacies to {{ location }}</h2>
```
st3v3nhunt: Should this class be empty?

app/views/results.nunjucks
```diff
@@ -9,34 +9,84 @@
...
+        <ol class="results__nearby-list">
         {% for service in nearbyServices %}
-        <li class="list-results-item nearby" style="background-color:cadetblue;">
-          {% include "includes/result-item.nunjucks" %}
-        </li>
+          <li class="results__item results__item--nearby">
```
st3v3nhunt: Could the `<li>` and `<div>` be moved into the `result-item` include, if they are being added to each item?

app/views/results.nunjucks
mcheung-nhs: Nope - should have "results__type results__type--nearby"

app/views/results.nunjucks
```diff
@@ -9,34 +9,84 @@
...
+      <ul>
+        <li>medical advice if you aren’t in an emergency</li>
+        <li>advice on whether you need urgent care</li>
+        <li>more information about a specific health condition</li>
+      </ul>
+      <p>If you need emergency help, call 999.</p>
+    </div>
+  </div>
+
   {% else %}
```
st3v3nhunt: This can go and the line 33 can be changed to `{% elif openServices.length > 0 %}` and line 87 can be removed - I think.

app/views/results.nunjucks
st3v3nhunt: Ignore this comment, I was looking at the two nearby blocks (which could be added into their own include) but due to the open list my original comment doesn't make sense!

PR #211, opened 2016-12-14
Fix/content

st3v3nhunt:


app/lib/postcodes.js
```diff
@@ -13,7 +13,8 @@ function lookup(res, next) {
   if (outcode) {
     url = `${baseUrl}/outcodes/${location}`;
```
st3v3nhunt: No need to run this through the encoder as there are no spaces in it

test/unit/middleware/setLocals.js
```diff
@@ -51,23 +38,14 @@ describe('setLocals', () => {
           },
           get: getNull,
         };
-        res = {
-          locals: {
-            backLink: {},
-          },
-        };
+        res = { locals: { backLink: {} } };
```
st3v3nhunt: Just a bit of a tidy up, trying to reduce the length of the file as much as anything

test/unit/middleware/setLocals.js
```diff
@@ -15,25 +15,12 @@ describe('setLocals', () => {
...
-          },
-          get: getNull,
-        };
-        res = {
-          locals: {
-            backLink: {},
-          },
-        };
+        req = { query: { }, get: getNull };
+        res = { locals: { backLink: {} } };
```
st3v3nhunt: Just a bit of a tidy up, trying to reduce the length of the file as much as anything

st3v3nhunt:
[New Pharmacy Results Page](https://trello.com/c/S3U6IZg8/611-new-pharmacy-results-page)

app/middleware/locationValidator.js
```diff
@@ -2,6 +2,13 @@ const log = require('../lib/logger');
 const renderer = require('../middleware/renderer');
 const locationValidator = require('../lib/locationValidator');
 
+function setLocationLabel(res, location) {
+  if (location) {
+    // eslint-disable-next-line no-param-reassign
+    res.locals.locationLabel = 'Enter a valid postcode';
```
neilbmclaughlin: Are we moving towards all text being defined in middleware? Just trying to work out why we have the page title hard coded in the view. In principle, it's a good idea but not sure of the benefit it brings us.

neilbmclaughlin: Are we moving towards all text being defined in middleware? Just trying to understand why we have the page title hard coded in the view but location label is defined in locals. In principle, it's a good idea but not sure of the benefit it brings us as I think it is currently only used in a single place.

st3v3nhunt: In response to the now deleted comment about why that text is there (a good question and something I should have commented on when I did it):

I've done it because it is variable (gets overridden when an invalid location has been entered) and I didn't want any logic for determining which text to use in the view.

Good idea to set the page title in middleware.

neilbmclaughlin: Not a blocker to approval just trying to understand.

PR #215, opened 2016-12-15
Feature/log zero results

st3v3nhunt:


app/middleware/getPharmacies.js
```diff
@@ -3,6 +3,14 @@ const log = require('../lib/logger');
 const dedupe = require('../lib/dedupePharmacies');
 const constants = require('../lib/constants');
 
+function logZeroResults(res) {
```
st3v3nhunt: Is this the best place to do the logging or do it as its' own middleware? I'm struggling to choose 🤔 
What do you reckon?

st3v3nhunt:
[Log searches returning zero results](https://trello.com/c/a2auauJS/618-log-searches-returning-zero-results)

app/middleware/getPharmacies.js
neilbmclaughlin: In terms of each bit of middleware doing _one thing well_ and _should only have one reason to change_ (™) it would make sense for it to be in it's own middleware - getting and logging are not the same. Not sure what the performance overhead of adding another middleware and if you decided not to I wouldn't argue.

st3v3nhunt:
I've opted to have the logging done via middleware.

I've added tests for it through the use of [`proxyquire`](https://www.npmjs.com/package/proxyquire). Other packages to do the same thing are available e.g. [`mock-require`](https://www.npmjs.com/package/mock-require) or [`sinon`](https://www.npmjs.com/package/sinon). I went with the former due to it being well used in comparison to `mock-require` and just having the features I needed rather than using `sinon` which has much more functionality.

PR #216, opened 2016-12-16
:innocent: Return no more than 3 nearby results

st3v3nhunt:


st3v3nhunt:
[Remove 4th nearby result if the open one is not one of them](https://trello.com/c/gxhJitJX/628-remove-4th-nearby-result-if-the-open-one-is-not-one-of-them)

test/unit/lib/dedupePharmacies.js
```diff
@@ -48,6 +48,28 @@ describe('dedupePharmacies', () => {
...
+          { identifier: 'AAA02' },
+          { identifier: 'AAA03' },
+          { identifier: fourthId },
+        ] };
+
+      const result = dedupe(services);
+
+      expect(result.open.length).is.equal(1);
+      expect(result.nearby.length).is.equal(3);
+      result.nearby.forEach((item) => {
```
neilbmclaughlin: There are collection expectations contains and includes which I'm guessing can be negated. (http://chaijs.com/api/bdd/)

st3v3nhunt: Care to provide an example?

neilbmclaughlin: Not at my machine but as a guess

expect(nearby.map((p) => p.identifier)).to.not.include(fourthId)

test/unit/lib/dedupePharmacies.js
```diff
@@ -4,47 +4,58 @@ const chai = require('chai');
...
+      };
 
       const result = dedupe(services);
 
       expect(result.open.length).is.equal(0);
       expect(result.nearby.length).is.equal(3);
-      result.nearby.forEach((item) => {
-        expect(item.identifier).to.not.equal(fourthId);
-      });
+      expect(result.nearby).to.deep.not.include.members([fourthItem]);
```
neilbmclaughlin: Just an observation but it's concise and does the job but it doesn't read well.

PR #222, opened 2017-01-10
:mouse: Add more detail to development instructions

st3v3nhunt:


ajthomascouk:
Do you think we need to be even more obvious and say they need to be in the cloned folder?

st3v3nhunt:
@ajthomascouk How about that? I've added a bit more detail and moved it closer to the top so should be closer to the first thing people read.

PR #224, opened 2017-01-11
Feature/sassify

ajthomascouk:


neilbmclaughlin:
[SASSIFY Everything](https://trello.com/c/1ZoV6W5N/557-sassify-everything)

ajthomascouk:
Using `compressed` CSS gives an error in codacy. Using `compacted` doesn't.

package.json
```diff
@@ -11,7 +11,8 @@
     "start-logging": "node app.js | ./node_modules/.bin/bunyan",
     "lint": "eslint --ext .js,.json .",
     "test": "NODE_ENV=test mocha --recursive test",
-    "watch-dev": "nodemon app.js | ./node_modules/.bin/bunyan",
+    "build-css": "./css-build",
+    "watch-dev": "nodemon app.js | ./node_modules/.bin/bunyan | ./css-watch",
```
st3v3nhunt: If you use a single ampersand rather than a pipe both processes will run and you will see output from them in the foreground. You can also get rid of the `css-watch` file and do it all in one command e.g. `nodemon app.js | ./node_modules/.bin/bunyan & nodemon -e scss -x npm run build-css`,

css-watch
```diff
@@ -0,0 +1 @@
+nodemon -e scss -x "npm run build-css"
```
st3v3nhunt: See comment in `package.json` about not needing this file

st3v3nhunt:
Somethings I've noticed:

1. The files checked into `public/css/*` look to be the compressed versions which means when I run node-sass they are all changed.
1. The last commit needs an emoji!
1. The cookie banner is not disappearing once it has been seen once.
1. The width of the opening time message is not as wide as it is on the current live site (see images)

## Current live:
screen shot 2017-01-13 at 14 14 25

## Changed:
screen shot 2017-01-13 at 14 13 58

package.json
ajthomascouk: Originally I had this all on one line, but wasn't sure if this was best practice. I'll go with your suggestion.

css-watch
ajthomascouk: Removed css-watch

ajthomascouk:
@neilbmclaughlin said:
> Where did the beta health scss files come from

https://github.com/nhsuk/betahealth/tree/develop/assets/stylesheets

> Also, Is there a plan for how we might up date them to reflect future changes.

Ideally we'd like to get notified of any changes to the files, whether that be from the Beta team or some GIT extension. @st3v3nhunt mentioned something that we could investigate, but the name escapes me.

> Just a couple of questions before I approve

I'd actually like to skip this branch, as well as feature/serve-css and feature/sha-css and focus on my latest branch, feature/sass-env (not pushed yet). Each branch has been branched from its predecessor and supersedes it. I'm not sure this was the right approach?

PR #225, opened 2017-01-12
:bug: Whitelist hotjar to get survey to work

st3v3nhunt:
Using `child-src` rather than `frame-src` due to `frame-src` having been [deprecated](https://www.w3.org/TR/CSP2/#directive-frame-src)

To confirm this is working (at least before it is deployed to a site where the survey will be active) navigate to the results page of the deployment and check there are no warnings in the browser's console about `Refused to frame 'https://vars.hotjar.com/rcj-hotjar-265857.js?sv=5:44` (or similar). To see what errors are occurring - check them out on the live or staging site.

### DIFF
--- a/config/express.js
+++ b/config/express.js
@@ -28,6 +28,9 @@ module.exports = (app, config) => {
       defaultSrc: [
         '\'self\'',
       ],
+      childSrc: [
+        'https://*.hotjar.com:*',
+      ],
       scriptSrc: [
         '\'self\'',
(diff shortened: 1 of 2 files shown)


st3v3nhunt:
[BUG: Hotjar's not working](https://trello.com/c/SZqHfDuz/681-bug-hotjar-s-not-working)

st3v3nhunt:
Confirmed the survey is showing on staging

PR #229, opened 2017-01-17
Feature/sha css

ajthomascouk:
Added two modules:
**hashmark** - Adds a hash of specified length to the css file(s) name.
**replaceinfiles** - Replaces original referenced files to new hashed files in a specified file(s)


ajthomascouk:
[Add SHA to outputted CSS file name](https://trello.com/c/jbUypooI/683-add-sha-to-outputted-css-file-name)

neilbmclaughlin:
The issue with when the CSS files are generated will be resolved as part of the Docker work.

ajthomascouk:
codacy always complains about the compressed CSS, can we do something about it?

teneightfive:
@ajthomascouk what's the reason for committing the compiled css files? It would be better to exclude those files from the repo and also do the hashing of the filenames to bust cache as part of the build process.

replace.json
```diff
@@ -0,0 +1,3 @@
+{
+  "/* @echo FONT_CDN_PATH *":"https://dhrlmnmyf2njb.cloudfront.net"
```
teneightfive: This should really be an environment variable rather than committing it into source control. We're using different values for staging and production which this method then supports.

Anything that has a reference to a path or an ID it's often useful to try and store them as a variable that can be configured differently depending on the environment the app is being run on.

neilbmclaughlin:
@teneightfive - there is a separate PR to exclude the generated css from the repo make the css build part of the build process.

neilbmclaughlin:
I've added the css directory to the ignore list in codacy so it should stop the comments. This is a known issue with codacy ([here](http://feedback.codacy.com/forums/279465-general/suggestions/11639442-support-scss-sass-files)). Unfortunately the current codacy review has hung on 'analysing ...' and I'm not sure how to force it to restart. Hopefully it will rerun on the next commit and all will be hunky dory.

neilbmclaughlin:
Sorry it's taken so long to do a proper review on this.

Our policy to date is to use npm to do everything unless there is a reason not to and this has meant that we haven't found a reason for adopting grunt/gulp/etc.

build-css, although called from npm, is a shell script which itself calls node modules.

It would be better to do this completely within node as it provides more portability and node is better at dealing with JSON objects and streams which is essentially what the 'sassify' problem is.

There is a sample of what it _might_ look like [here](https://gist.github.com/neilbmclaughlin/03fc7ed0b750c54053d187eb3dc01135). The sample doesn't consider the token replacement issue but that is a stream transform problem which node could handle too.

Sorry to take so long to get to this conclusion despite having looked at the code a few times - I hadn't appreciated that `npm run build-css` was actually running a shell script. I'm happy to pair on Monday to make the changes.

replace.json
ajthomascouk: Tried to get this working using an environment variable. @teneightfive, maybe we could pair on this?

ajthomascouk:
>  I'm happy to pair on Monday to make the changes.
Yeah, the bash script wasn't ideal, so a pairing session would be good @neilbmclaughlin, thanks.

neilbmclaughlin:
There is an argument to say that we should be using a per file hash value. At the moment with the timestamp approach every rebuild will bust the cache even if the css hasn't actually changed. Something to consider.

neilbmclaughlin:
Testing closing and opening the PR to see if resolves the hung codacy analysis problem.

replace.json
teneightfive: @ajthomascouk Yep, happy to pair to solve this

teneightfive:
@neilbmclaughlin yep, I agree using a timestamp per build isn't ideal. We only really want to bust the cache for assets if they contents change. That's what we're currently doing) in betahealth. We're using gulp though which makes including modules to handle things like revisioning assets easier.

There's two parts to this though:

1. [Revision the assets](https://github.com/nhsuk/betahealth/blob/develop/gulp/tasks/shared/rev.js) based on their contents and create a manifest file
1. Use [middleware](https://github.com/nhsuk/betahealth/blob/develop/app/middleware/asset-path.js) within express to check the path needing to be served against the manifest and output the revisioned filepath

neilbmclaughlin:
We are going to park this work for now pending the work to have shared assets.

PR #231, opened 2017-01-23
Don't build Docker images for commits

adamgraves-choices:
Also replaces special characters in branch name, so docker can tag and push the branch.


neilbmclaughlin:
[Change deployment mechanism from Azure WebApps to Docker](https://trello.com/c/F4ySZOIz/685-change-deployment-mechanism-from-azure-webapps-to-docker)

adamgraves-choices:
@neilbmclaughlin after discussion with @st3v3nhunt we feel that there's no need to build either commit IDs or branch names, on the basis that branch if a branch needs reviewed by others it should already be a PR.

I've kept the tagging of the master branch initially.

PR #241, opened 2017-02-01
Generate css as part of docker image build

neilbmclaughlin:
Note that this PR contains work on the rancher deployment per pull request as well as the scss/css image building. This is deliberate as we were using a relatively minor front end change as a test bed for the rancher deployment scripts.


scripts/rancher_pr_deploy.sh
```diff
@@ -24,13 +24,15 @@ if [[ -n "$TRAVIS" ]]; then
     curl -s https://raw.githubusercontent.com/nhsuk/nhsuk-rancher-templates/master/templates/c2s-pharmacy-finder/0/docker-compose.yml  -o docker-compose.yml
     curl -s https://raw.githubusercontent.com/nhsuk/nhsuk-rancher-templates/master/templates/c2s-pharmacy-finder/0/rancher-compose.yml -o rancher-compose.yml
 
+    LATEST_RELEASE=$(curl -s https://api.github.com/repos/nhsuk/nearby-services-api/releases/latest | jq -r '.tag_name')
```
neilbmclaughlin: Note that this change assumes that the pre-release option has been deselected in GitHub for the current release. See [here](https://developer.github.com/v3/repos/releases/#get-the-latest-release) for more details. It also assumes that there is an image associated with the release - I think these have been created manually so far.

neilbmclaughlin:
Fonts are broken. Probably as we are calling from the wrong domain.

adamgraves-choices:
@neilbmclaughlin 

That's correct. See https://c2s-production.trafficmanager.net/finders/find-help vs https://beta.nhs.uk/finders/find-help. Same endpoint.

scripts/rancher_pr_deploy.sh
neilbmclaughlin: Options for creating release and images:

- Script the creation of a release so we can also generate the tagged image at the same time. See [here](https://developer.github.com/v3/repos/releases/#create-a-release) for more details. Note: we should mark it as pre-release until we are ready to start using it.
- Alternative would be to use the GitHub release webhook

In both cases, we need somewhere to run the script/webhook. (?Azure Functions vs Docker Container)

neilbmclaughlin:
Could do with a webhook which posts the URL for the deployed PR to the GitHub comments. Also, PR deployment should be the same each time (i.e. not change name each time there is a new build of the PR).

neilbmclaughlin:
If this passes review we can disable Heroku integration.

Dockerfile
```diff
@@ -15,4 +15,6 @@ EXPOSE 3000
 
 COPY . /code
 
+RUN [ "npm", "run", "build-css" ]
```
st3v3nhunt: Could/should this be done as part of `npm start` i.e. change `package.json` to be `"start": "npm run build-css && node app.js"`

neilbmclaughlin: As discussed in the c2s-dev catch up this will stay as is for now but the read me has been updated to reflect the need to run build-css when running the app locally.

PR #242, opened 2017-02-02
Rancher environment var has been added to travis

adamgraves-choices:


adamgraves-choices:
The main fix was adding a set of ENV variables for rancher to the Travis build. See the #c2s-dev chat. I've fixed the json payload so the slack notifications are working now, that posts to #c2s-notice.

PR #245, opened 2017-02-03
:package: Run tests in container

st3v3nhunt:
Instead of running tests on the CI's host OS we should be running them inside of the container. Using the `git-hook` script, linting is done, tests are run, code coverage generated and uploaded to both coveralls and codacy.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,16 +3,9 @@ sudo: required
 services:
   - docker
 
-language: node_js
-
-node_js:
-- node
 script:
-- npm run lint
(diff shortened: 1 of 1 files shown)


st3v3nhunt:
@beseven & @neilbmclaughlin A relatively small change here but worth noting - the testing of the code is done within the containers on CI now. The reason being to have the tests run on the exact OS that will be running in all other environments. Instead of how it works atm where the tests run on different OSes between dev, CI and deployed environments.

PR #274, opened 2017-03-21
Feature/non root user

st3v3nhunt:


c2s-dev:
:rocket: deployed to [http://connecting-to-services-pr-274.dev.c2s.nhschoices.net](http://connecting-to-services-pr-274.dev.c2s.nhschoices.net)

PR #277, opened 2017-03-30
:palm_tree: Log as object

st3v3nhunt:


c2s-dev:
:rocket: deployed to [http://connecting-to-services-pr-277.dev.c2s.nhschoices.net](http://connecting-to-services-pr-277.dev.c2s.nhschoices.net)

PR #280, opened 2017-04-18
Bug/isle of man

markysoft:
Fix for Isle of map postcodes not returning lat/longs and causing error


c2s-dev:
:rocket: deployed to [http://connecting-to-services-pr-280.dev.c2s.nhschoices.net](http://connecting-to-services-pr-280.dev.c2s.nhschoices.net)

c2s-dev:
:rocket: deployed to [http://connecting-to-services-pr-280.dev.c2s.nhschoices.net](http://connecting-to-services-pr-280.dev.c2s.nhschoices.net)

c2s-dev:
:rocket: deployment of connecting-to-services-pr-280 in nhsuk-dev succeeded (http://connecting-to-services-pr-280.dev.beta.nhschoices.net)

st3v3nhunt:
@markysoft Can you rebase the branch please so the changes for the deployment to the new domain are included?

st3v3nhunt:
I wonder whether it might be better to check if the (post|out)code is an Isle of Man location. If it is, direct people straight to the no results found page. Or add another page specifically for the IoM. Currently an IMx outcode breaks this implementation.

markysoft:
Interesting point, we could also do the same for Wales and Scotland, could suggest it as an enhancement for another card.

st3v3nhunt:
Possibly but then you have the issue of restricting the results just because the search is 'over the border'. That might be where they are physically but it could be within the search distance of a pharmacy that we have data on. Previously, the way we have dealt with that is to allow for the results to be displayed even when people are 'over the border'. I doubt that would change for this. There are also some postcodes that aren't actually in the country you might think they are in. Which is why the outcode checking hasn't been employed in the past - it just isn't reliable.

st3v3nhunt:
I think we have the same breaking bug for the Channel Islands:

* [Guernsey, Channel Islands](https://beta.nhs.uk/finders/results?location=GY1)
* [Jersey, Channel Islands](https://beta.nhs.uk/finders/results?location=JE1)

c2s-dev:
:rocket: deployment of connecting-to-services-pr-280 in nhsuk-dev succeeded (http://connecting-to-services-pr-280.dev.beta.nhschoices.net)

c2s-dev:
:rocket: deployment of connecting-to-services-pr-280 in nhsuk-dev succeeded (http://connecting-to-services-pr-280.dev.beta.nhschoices.net)

c2s-dev:
:rocket: deployment of connecting-to-services-pr-280 in nhsuk-dev succeeded (http://connecting-to-services-pr-280.dev.beta.nhschoices.net)

app/views/includes/no-results.nunjucks
```diff
@@ -3,10 +3,11 @@
     <h2 class="results__header results__header--none">There are no pharmacies within 20 miles of {{ location }}</h2>
     <div class="results__none-content">
       <p>This service only provides information about pharmacies in England.</p>
-      <p>You might be getting this result because you live in Scotland, Wales or Northern Ireland, or because you live near a border with one of those countries.</p>
-      <p><a href="http://www.nhs24.com/findlocal/">Find pharmacies in Scotland on the NHS 24 website.</a></p>
-      <p><a href="http://www.nhsdirect.wales.nhs.uk/localservices/searchlocalservices.aspx?s=Pharmacy">Find pharmacies in Wales on the NHS Direct Wales website.</a></p>
-      <p><a href="http://servicefinder.hscni.net">Find pharmacies in Northern Ireland on the Health and Social Care Online website.</a></p>
+      <p>You might be getting this result because you live in Scotland, Wales or Northern Ireland, or because you live near a border with one of those countries. It could also be because you live on the Isle of Man.</p>
```
beseven: Can you add the Channel Islands too?

markysoft:
I'll ask Dean for input on new wording.

It's unclear as to how I would move the post/outcode checks into the validateLocation, as this is concerned with checking whether a postcode/outcode is valid, whereas the new logic is for adding missing location details to the results from a valid postcode/outcode lookup.

st3v3nhunt:
@markysoft Checking whether the (post|out)code is valid is what needs to happen here though? Those are not valid for this search.
I don't agree with the approach of adding the lat/lon. It happens surreptitiously and doesn't help get any results either (as there are no locations on the islands, at least that we have data for).

markysoft:
I'd be inclined to scrap the whole approach and decide on an error message when there is no lat long

c2s-dev:
:rocket: deployment of connecting-to-services-pr-280 in nhsuk-dev succeeded (http://connecting-to-services-pr-280.dev.beta.nhschoices.net)

test/unit/lib/isNotEnglishLocation.js
```diff
@@ -0,0 +1,72 @@
+const isNotEnglishLocation = require('../../../app/lib/isNotEnglishLocation');
+const chai = require('chai');
+
+const expect = chai.expect;
+
+describe('isNotEnglishLocation', () => {
```
st3v3nhunt: Has the test for `IM` postcode and one ending in `IM` been purposefully omitted?

app/middleware/locationValidator.js
```diff
@@ -30,4 +31,20 @@ function validateLocation(req, res, next) {
...
+  /* eslint-disable no-param-reassign */
+  res.locals.nearbyServices = [];
+  res.locals.openServices = [];
+  /* eslint-enable no-param-reassign*/
+  renderer.results(req, res);
+}
+
+function validateLocation(req, res, next) {
+  if (isNotEnglishLocation(res.locals.location)) {
+    renderNoResultsPage(req, res);
```
st3v3nhunt: Can a log statement of what the postcode was be added please? It would be good to have a record of what has been added.

app/middleware/locationValidator.js
```diff
@@ -30,4 +31,20 @@ function validateLocation(req, res, next) {
   }
 }
 
+function renderNoResultsPage(req, res) {
+  /* eslint-disable no-param-reassign */
+  res.locals.nearbyServices = [];
```
st3v3nhunt: I don't think there is any need to set either `nearbyServices` or `openServices` to empty arrays.

markysoft: The error message doesn't show in the view if these are not set.

c2s-dev:
:rocket: deployment of connecting-to-services-pr-280 in nhsuk-dev succeeded (http://connecting-to-services-pr-280.dev.beta.nhschoices.net)

app/views/includes/no-results.nunjucks
st3v3nhunt: No wording changes will be made as part of this change. For a future iteration...

st3v3nhunt:
@beseven Can you review and merge if you are happy please?

PR #304, opened 2017-07-03
Fix Splunk tagging

adamgraves-choices:
See: http://rancher.com/docs/rancher/v1.5/en/cli/variable-interpolation/#escaping-double-brackets

### DIFF
--- a/rancher-config/docker-compose.yml
+++ b/rancher-config/docker-compose.yml
@@ -24,7 +24,7 @@ services:
         splunk-token: ${SPLUNK_HEC_TOKEN}
         splunk-insecureskipverify: "true"
         splunk-sourcetype: docker
-        splunk-source: pharmacy-finder
+        splunk-source: connecting-to-services
         splunk-verify-connection: "false"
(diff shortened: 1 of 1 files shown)


c2s-dev:
:rocket: deployment of connecting-to-services succeeded (http://connecting-to-services-pr-304.dev.beta.nhschoices.net)

rancher-config/docker-compose.yml
```diff
@@ -26,5 +26,5 @@ services:
         splunk-sourcetype: docker
         splunk-source: pharmacy-finder
```
st3v3nhunt: I know this hasn't been changed but do you think rather than `pharmacy-finder` the source should be `connecting-to-services`?

adamgraves-choices: I was just thinking about that. Feel like we should maybe rename the repo instead to `pharmacy-finder`?

c2s-dev:
:rocket: deployment of connecting-to-services succeeded (http://connecting-to-services-pr-304.dev.beta.nhschoices.net)

PR #307, opened 2017-07-06
:lips: No need to install git in container

st3v3nhunt:


c2s-dev:
:rocket: deployment of connecting-to-services succeeded (http://connecting-to-services-pr-307.dev.beta.nhschoices.net)

PR #353, opened 2017-10-09
:no_bell: switch performance test threshold to average

markysoft:
Add a threshold file to specify the minimum average speed for a test run, rather than compare the time to the previous run. 
## Checklist

- [ ] Version of the cache is increased


markysoft:
[Update all apps to have perfomance test run on a schedule against master](https://trello.com/c/OiKGroSr/908-update-all-apps-to-have-perfomance-test-run-on-a-schedule-against-master)

CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+0.14.1 / TBA
+- Performance tests use threshold of 750ms
```
st3v3nhunt: Missing underline.

beseven:
I'll ask the same question that @st3v3nhunt asked me, if you put it down to 500ms but tweak the throuput that it has, will it not achieve what you want?

markysoft:
Changing the throughput would allow us to set 500ms as the target, but the current throughput is meant to simulate typical load on the server. 

I suppose it depends on the goal of the tests. We can dial the throughput down so we can get reliable response times that simulate quiet periods, or use a realistic figure with some headroom in the response time.

PR #355, opened 2017-10-11
Remove caching

markysoft:
## Description
Remove smart cache.

## Motivation and Context

Search results for pharmacies were stale in production as they were being cached for 30 minutes.

## Checklist


CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+0.13.1 / 2017-10-11
```
st3v3nhunt: The removal of the cache feel more like a feature than a bug fix to me. If I was adding it I would certainly be thinking it would be a minor version increase as opposed to a patch version.

rancher-config/docker-compose.yml
```diff
@@ -10,7 +10,7 @@ services:
       GOOGLE_ANALYTICS_TRACKING_ID: ${GOOGLE_ID}
       WEBTRENDS_ANALYTICS_TRACKING_ID: ${WEBTRENDS_ANALYTICS_TRACKING_ID}
       API_BASE_URL: 'http://nearby-services-api.nearby-services-api:3001'
-      CACHE_TIMEOUT_SECONDS: 1800
+      CACHE_TIMEOUT_SECONDS: 0
```
st3v3nhunt: As this is no longer used can it be removed?

CHANGELOG.md
markysoft: Agreed. Have upped the minor version instead

PR #356, opened 2017-10-13
:raising_hand: Be less specific about third party domains

st3v3nhunt:
Reason for being less specific on the domains is due to occassional
changes by the third parties which causes functionality to stop working
on our site.

Removal of scheme specification as it defaults to how the page has been
accessed.

### DIFF
--- a/config/express.js
+++ b/config/express.js
@@ -48,7 +48,6 @@ module.exports = (app, config) => {
           '*.hotjar.com',
           '*.webtrends.com',
           '*.webtrendslive.com',
-          'cdn.jsdelivr.net',
         ],
         imgSrc: [
           '\'self\'',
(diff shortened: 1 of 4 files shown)


config/express.js
```diff
@@ -28,55 +28,50 @@ module.exports = (app, config) => {
...
-        '\'self\'',
-        '\'unsafe-inline\'',
-        '\'unsafe-eval\'',
-        'data:',
-        'www.google-analytics.com',
-        's.webtrends.com',
-        'statse.webtrendslive.com',
-        'static.hotjar.com',
-        'script.hotjar.com',
-        'cdn.jsdelivr.net',
```
markysoft: is `cdn.jsdelivr.net` no longer used?

st3v3nhunt: I'm not sure it was ever used and if it was there are no requests from any of the apps going to that domain now.

PR #364, opened 2017-10-26
:closed_lock_with_key: Add basic authentication to test environments in Rancher

st3v3nhunt:
Adds basic auth to the service. Controlled via vault variables. By default all environments are protected. Overridden for environments that require access by all.

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,8 +25,14 @@ Use the `test` script for continuous testing during development.
 ## Test environments
 
 As the application is being developed, every Pull Request has its own test
-environment automatically built and deployed to. The path to the environment
-is added as a comment to the PR in GitHub.
+environment automatically built and deployed to.
+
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -28,6 +28,13 @@ As the application is being developed, every Pull Request has its own test
 environment automatically built and deployed to. The path to the environment
 is added as a comment to the PR in GitHub.
```
neilbmclaughlin: This is no longer true - the link is in the checks (as the Details for rancher/deployment check).

The readme has not been updated to reflect this change.

st3v3nhunt: Add as an issue.

st3v3nhunt: My bad. Sorted this and alphabetised the docker config to make it easier to search through.

PR #369, opened 2017-10-31
search by place

markysoft:
## Description
Add support for search by place

## Checklist


app/lib/sortByLocalType.js
```diff
@@ -0,0 +1,24 @@
+const localOrder = {
+  City: 5,
```
st3v3nhunt: What is the likelihood these will change?
If they did change or another one was added the comparisons on lines [14](https://github.com/nhsuk/connecting-to-services/pull/369/files#diff-6f5c92439fec630a69c6d2ecf8cb158fR14) and [17](https://github.com/nhsuk/connecting-to-services/pull/369/files#diff-6f5c92439fec630a69c6d2ecf8cb158fR17) are going to end up making that new `local_type` be smaller than anything else. And we aren't going to know about it.
It would be good to have something logged so that we could alert on such an event happening.

app/middleware/skipLatLongLookup.js
```diff
@@ -0,0 +1,5 @@
+function skipLatLongLookup(res) {
```
st3v3nhunt: Rather than using this technique to conditionally execute route functions I was thinking it would be possible to use pattern matching within the route definitions to capture the query string params and route based on that. See https://expressjs.com/en/guide/routing.html#route-parameters

app/middleware/locationValidator.js
```diff
@@ -9,37 +12,56 @@ function setLocationLabel(res, location) {
...
-    res.locals.errorMessage = validationResult.errorMessage;
-    setLocationLabel(res, location);
-    renderer.findHelp(req, res);
+    renderFindHelpPage(req, res, location, 'Non-English postcode', validationResult.errorMessage);
   } else {
     next();
   }
 }
+function sanitiseString(string) {
+  return string.replace(/[^a-z]/gmi, ' ').replace(/\s\s+/g, ' ').trim();
```
st3v3nhunt: This looks very similar to https://github.com/nhsuk/connecting-to-services/pull/369/files#diff-e675aab0ca0472369ec07b2b8552e5a9R13. Perhaps they can be extracted into a utility method?

app/middleware/createPlaceViewModel.js
```diff
@@ -0,0 +1,14 @@
+function createPlaceViewModel(place) {
+  return {
+    name: place.name_1,
```
st3v3nhunt: Does `name_1` need to be URL encoded?

app/middleware/getPlaces.js
```diff
@@ -0,0 +1,43 @@
+const log = require('../lib/logger');
+const locate = require('../lib/locate');
+const sortByLocalType = require('../lib/sortByLocalType');
+const renderer = require('./renderer');
+const placeSearches = require('../lib/promCounters').placeSearches;
+const placeDisambiguationViews = require('../lib/promCounters').placeDisambiguationViews;
+
+// to do move to common library
```
st3v3nhunt: I assume this is going to go at some point?

app/middleware/getPlaces.js
```diff
@@ -0,0 +1,43 @@
...
+function renderFindHelpPage(req, res, location, message, errorMessage) {
+  log.info({ locationValidationResponse: { location } }, message);
+  res.locals.errorMessage = errorMessage;
+  renderer.findHelp(req, res);
+}
+
+function logZeroResults(places, location) {
+  if (places.length === 0) {
+    log.warn({ location }, `No results were found for ${location}`);
```
st3v3nhunt: Is it worth recording the number of times this happens with Prometheus too?

app/middleware/getPlaces.js
```diff
@@ -0,0 +1,43 @@
...
+  }
+}
+
+function getLocationUrl(place, location) {
+  return `results?location=${location}&latitude=${place.latitude}&longitude=${place.longitude}`;
```
st3v3nhunt: Does `location` need URL encoding?

app/views/places.nunjucks
```diff
@@ -0,0 +1,38 @@
...
+<div class="grid-row">
+
+  <h1 class="sr-only">Search results</h1>
+
+  {% if places.length === 0 %}
+
+    {% include "includes/no-place-results.nunjucks" %}
+
+  {% elif places.length > 0 %}
+  <h2 class="results__header">
```
st3v3nhunt: Is it worth pulling this section out into an include?

doc/adr/0009-search-by-place.md
```diff
@@ -0,0 +1,19 @@
...
+
+Date: 2017-11-03
+
+## Status
+
+Accepted
+
+## Context
+
+Pharmacies may only be located by postcode or outcode.
```
st3v3nhunt: I'd go with something along the lines of `There is a need to provide the ability for users to search by place name.` for the context.

test/integration/locate.js
```diff
@@ -0,0 +1,62 @@
+const chai = require('chai');
+const lookup = require('../../app/lib/locate');
+
+const expect = chai.expect;
+
+describe('locate', () => {
+  describe('byPostcode', () => {
+    it('Should return lat long for valid England postcode', async () => {
```
st3v3nhunt: Capitalised `Should` goes against the convention in the other tests.

test/integration/locate.js
```diff
@@ -0,0 +1,62 @@
...
+      const validPostcode = 'BD24 9PT';
+      const result = await lookup.byPostcode(validPostcode);
+
+      expect(result.postcode).to.equal(validPostcode);
+      /* eslint-disable no-unused-expressions */
```
st3v3nhunt: See comment regarding the unit test specific eslint config which, if added would mean these comments could be removed.

test/integration/locate.js
```diff
@@ -0,0 +1,62 @@
+const chai = require('chai');
+const lookup = require('../../app/lib/locate');
+
+const expect = chai.expect;
+
+describe('locate', () => {
```
st3v3nhunt: With these tests using the actual service they aren't atomic. Given `locate.js` doesn't do much beyond use the `Postcodes.io` library to do the lookup I'd suggest these tests are removed and the utility functions in `locate.js` are moved into a utils method and tested specifically.

app/lib/sortByLocalType.js
markysoft: I'll log an error on 'undefined' from the lookup to ensure we are aware of new codes being introduced.

app/middleware/getPlaces.js
markysoft: this will be removed when combine the routes

test/integration/locate.js
markysoft: Is the stance that we do not write any integration tests that call external services? Without this test all calls will be mocked and we won't be alerted to any breaking changes to postcodes.io

st3v3nhunt: Given this is the exact same service all of the deployments of the site are using, including Beta/Live, I'd like to think we'd know the service had changed before these failed. And whilst I don't believe having these tests will help identify changes to the external service causing problems, having a test calling the service for a place unlikely to change e.g. `London` and testing the properties are OK is probably a good  idea from the POV of showing what values those properties have.

Having a test to confirm the integration of the service has been done correctly is valuable, however, I'd do that by going through the full stack and not just through the middleware. Ideally, it would also be run post-deployment too.

app/middleware/skipLatLongLookup.js
markysoft: I've looked into this, and there is no elegant way of routing on query strings other than using a regex.
The recent changes combine the place route into the results route uses the same conditionals. It might be worth getting together to look at it tomorrow, the current implementation isn't great - maybe we can find a way of improving the flow.

app/middleware/performPlaceSearch.js
```diff
@@ -0,0 +1,45 @@
...
+    if (places.length === 1) {
+      res.locals.coordinates = getCoordinates(places[0]);
+      next();
+    } else {
+      placeDisambiguationViews.inc(1);
+      res.locals.places = createPlaceViewModel(places);
+      renderer.places(req, res);
+    }
+  } catch (ex) {
+    routeHelper.renderFindHelpPage(req, res, location, 'Find places error');
```
st3v3nhunt: Should this render the find help page or an error page?
Looking at the equivalent for postcode lookup an error page would be returned.
The error gets logged too https://github.com/nhsuk/connecting-to-services/pull/369/files#diff-d2462ba736f50bc9398bf5779575bce9R23

app/middleware/routeHelper.js
```diff
@@ -0,0 +1,27 @@
+
```
st3v3nhunt: No need for this empty line

app/views/places.nunjucks
```diff
@@ -0,0 +1,26 @@
...
```
st3v3nhunt: The `h2` could probably go into the include.

app/middleware/locationValidator.js
```diff
@@ -1,45 +1,49 @@
...
-
-function setLocationLabel(res, location) {
-  if (location) {
-    res.locals.locationLabel = 'Enter a valid postcode';
-  }
-}
+const performPlaceSearch = require('./performPlaceSearch');
+const removeNonAlphabeticAndWhitespace = require('../lib/stringUtils').removeNonAlphabeticAndWhitespace;
 
 function validateEnglishLocation(req, res, next) {
```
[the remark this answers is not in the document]
markysoft: I'll rename validateEnglishLocation to validateEnglishPostcode

app/middleware/performPlaceSearch.js
markysoft: Changed to log an error and return the error page.

app/lib/sortByLocalType.js
st3v3nhunt: I think the order of these needs tweaking. See this [codelist](https://www.ordnancesurvey.co.uk/xml/codelists/LocalType.xml). Moving `Surburban Area` down to the lowest level would mean it gets sorted below the more easily well defined places and would be sorted alongside the other area type value of `Other Settlement`.

app/lib/sortByLocalType.js
```diff
@@ -0,0 +1,35 @@
...
+  const order = localOrder[type];
+  if (order === undefined) {
+    log.error(`unknown local_type '${type}'`);
+    return -1;
+  }
+  return order;
+}
+
+function sortByLocalType(places) {
+  return places.sort((a, b) => {
```
st3v3nhunt: In order to improve the consistency of the order of places sorting a second time alphabetically would be helpful. What do you reckon?

app/lib/sortPlace.js
```diff
@@ -18,18 +18,22 @@ function getOrder(type) {
...
     const first = getOrder(a.local_type);
     const second = getOrder(b.local_type);
-    if (first > second) {
-      return -1;
-    }
-    if (first < second) {
-      return 1;
-    }
-    return 0;
+    return compare(first, second) || compare(b.name_1, a.name_1);
```
st3v3nhunt: Can this be extended to include region and postcode please?

PR #372, opened 2017-11-02
:curly_loop: Make url more specific to finding a pharmacy

st3v3nhunt:


rancher-config/rancher-compose.yml
```diff
@@ -13,5 +13,5 @@ services:
       initializing_timeout: 60000
       interval: 5000
       strategy: recreate
-      request_line: GET "/finders/find-help" "HTTP/1.0"
+      request_line: GET "/find-a-pharmacy" "HTTP/1.0"
```
st3v3nhunt: Add trailing slash

PR #376, opened 2017-11-10
Add brunch and update deps

st3v3nhunt:


app/public/js/cookieMessage.js
```diff
@@ -0,0 +1,49 @@
+(function(global) {
+  'use strict'
+  var $ = global.jQuery;
+  var document = global.document;
+
+  function setCookie(name, value, options = {}) {
+    let cookieString = `${name}=${value}; path=/`;
```
markysoft: I realise this code isn't new, but are `let` and string templates compatible with the minimum browser versions we support?

st3v3nhunt: Its been cross browser tested so I would have thought so. However, the cookie banner is easy to miss so perhaps it doesn't work.

st3v3nhunt: Confirmed this needs to work on IE 7+ so needs to be `var`.
Source https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/let#Browser_compatibility

st3v3nhunt:
@markysoft Changes made.

st3v3nhunt:
@markysoft Additional IE 7 backwards compatibility changes made.

PR #382, opened 2017-11-14
:bowtie: content change: search message, quotes, disambiguation

markysoft:
`Enter a place or a postcode` on initial search screen is now `Enter a town, city or postcode`.
If a term is not found in a search, the search term is now surrounded by single quotes on the error page.
`There are n places matching x` is now `We found n places that match 'y'`.
`Please select your location from the list below to continue.` is now `Choose one of the following:`. The text has moved from the header into a styled `p` tag.
The search term on the results page is the cleansed search term rather than a string built of the address returned from the lookup.

### DIFF
--- a/app/lib/setLocationsOnLocals.js
+++ b/app/lib/setLocationsOnLocals.js
@@ -1,6 +1,0 @@
-function setLocationsOnLocals(res, location) {
-  res.locals.location = location;
-  res.locals.displayLocation = location && location.split(',')[0];
-}
-
-module.exports = setLocationsOnLocals;
--- a/app/lib/setLocationsOnLocals.js
+++ b/app/lib/setLocationsOnLocals.js
@@ -1,0 +1,6 @@
+function setLocationsOnLocals(res, location) {
+  res.locals.location = location;
+  res.locals.displayLocation = location && location.split(',')[0];
+}
+
+module.exports = setLocationsOnLocals;
--- a/app/views/includes/place-results.nunjucks
+++ b/app/views/includes/place-results.nunjucks
@@ -1,4 +1,5 @@
-
+<div class="reading-width places">
+  <h1 class="sr-only">Search results</h1>
   <h2 class="results__header">
     We found {{places.length}} places that match '{{location}}'
   </h2>
@@ -13,3 +14,4 @@
     </li>
   {% endfor %}
   </ol>
+  </div>
--- a/app/views/includes/place-results.nunjucks
+++ b/app/views/includes/place-results.nunjucks
@@ -1,7 +1,9 @@
 
   <h2 class="results__header">
-    There are {{places.length}} places matching {{location}}. <br/>
-    Please select your location from the list below to continue.
+    We found {{places.length}} places that match '{{location}}'
+  </h2>
+  <h2 class="results__header">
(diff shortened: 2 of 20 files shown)


app/views/includes/place-results.nunjucks
```diff
@@ -1,8 +1,10 @@
 
   <h2 class="results__header">
-    There are {{places.length}} places matching {{location}}. <br/>
-    Please select your location from the list below to continue.
+    We found {{places.length}} places that match '{{location}}'
   </h2>
+  <p>
```
st3v3nhunt: Does there need to be additional whitespace around the text?

app/views/includes/place-results.nunjucks
```diff
@@ -1,8 +1,10 @@
 
   <h2 class="results__header">
-    There are {{places.length}} places matching {{location}}. <br/>
-    Please select your location from the list below to continue.
+    We found {{places.length}} places that match '{{location}}'
```
st3v3nhunt: Could the whitespace be removed to have this on a single line?

test/integration/resultsPagePlace.js
```diff
@@ -80,7 +80,7 @@ describe('The place results page', () => {
         const $ = cheerio.load(res.text);
 
         expect($('.results__header').text())
-          .to.include('There are 3 places matching multiresult');
+          .to.include('We found 3 places that match \'multiresult\'');
```
st3v3nhunt: Is it worth extracting `multiresult` to a variable? It is used 3 times within this test.

test/integration/resultsPagePlace.js
```diff
@@ -121,7 +121,7 @@ describe('The place results page', () => {
       .end((err, res) => {
         iExpect.htmlWith200Status(err, res);
         const $ = cheerio.load(res.text);
-        expect($('.results__header--none').text()).to.be.equal('There are no matches for place noresults');
+        expect($('.results__header--none').text()).to.be.equal('We can\'t find \'noresults\'');
```
st3v3nhunt: Extract `noresults` to a variable as it is used 3 times?

app/lib/setLocationsOnLocals.js
```diff
@@ -0,0 +1,6 @@
+function setLocationsOnLocals(res, location) {
+  res.locals.location = location;
```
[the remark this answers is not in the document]
markysoft: A little unclear on this, are you suggesting we store the original location from the request on locals unaltered as`req_location` but keep the `location` that we can update?

As a bit of background the function was added to be called from all places where location is changed to ensure the displayLocation is kept in sync.

I think it's worth keeping the original unchanged as we made need it in a later feature.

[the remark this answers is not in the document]
st3v3nhunt: Sorry for the opaqueness of the comment.

I do think it would be worth keeping the original request value of location unaltered. Whether that is set via `res.locals.req = req.query` to get the full set of query string values or specifically via `res.locals.req_location = req.query.location`. I'd probably lean towards the latter as I'd imagine the former is a good attack vector for buffer overflows.

I wondered whether `res.locals.displayLocation` could be set to a function rather than a value to get around the need to set it more than once?

PR #387, opened 2017-11-17
:metal: add link to choices overview page

ajthomascouk:


app/views/includes/result-item.nunjucks
```diff
@@ -1,6 +1,7 @@
 <p>{{ service.distanceInMiles | round(1) }} miles away</p>
 <h2 class="results__name">{{ service.name }}</h2>
 <section class="callout callout--info callout--compact"><p class="callout--info__mesg">{{ service.openingTimesMessage }}</p></section>
+<p class="overview"><a href="{{ service.overviewUrl }}">See opening times and service<span class="sr-only">for {{ service.name }}</span></a></p>
```
st3v3nhunt: `service` needs to be pluralised i.e. `services`

app/lib/choicesOverview.js
```diff
@@ -0,0 +1,14 @@
+function choicesOverviewUrl(inputList) {
```
st3v3nhunt: Given the module name is `choicesOverview` I'd suggest changing the function's name to `addUrl` or something similar to describes what it is doing.

This function needs some unit tests.

app/lib/choicesOverview.js
```diff
@@ -0,0 +1,14 @@
...
+  return inputList.map((item) => {
+    const identifier = item.identifier;
+    const overviewUrl = `https://www.nhs.uk/Services/pharmacies/Overview/DefaultView.aspx?id=${identifier}`;
+
+    // eslint-disable-next-line no-param-reassign
+    item.overviewUrl = overviewUrl;
```
st3v3nhunt: I think `choicesOverviewUrl` would make it clearer as to what the URL was for. `overviewUrl` could be misinterpreted as an internal link within this application.

test/integration/resultsPagePlace.js
```diff
@@ -57,9 +57,15 @@ describe('The place results page', () => {
           expect($(link).attr('href')).to.have.string('https://maps.google.com');
         });
 
+        const overviewLinks = $('.overview a');
+        overviewLinks.toArray().forEach((link) => {
+          expect($(link).attr('href')).to.have.string('https://www.nhs.uk/Services/pharmacies/Overview/DefaultView.aspx');
```
st3v3nhunt: Assuming the (to be added) unit tests are confirming the correct id has been appended to the base url, it would be good to extend the test to confirm the number of links i.e. 4 have been displayed on the page.
It wouldn't hurt to check the text of the link either.

app/lib/choicesOverview.js
ajthomascouk: @st3v3nhunt wrote:
> Given the module name is choicesOverview I'd suggest changing the function's name to addUrl or something similar to describes what it is doing.
Done

app/lib/choicesOverview.js
ajthomascouk: Done

test/integration/resultsPagePlace.js
ajthomascouk: Done

test/integration/resultsPagePlace.js
```diff
@@ -57,6 +57,12 @@ describe('The place results page', () => {
           expect($(link).attr('href')).to.have.string('https://maps.google.com');
         });
 
+        const ChoicesOverviewLinks = $('.overview a');
+        ChoicesOverviewLinks.toArray().forEach((link) => {
+          expect($(link).attr('href')).to.have.string('https://www.nhs.uk/Services/pharmacies/Overview/DefaultView.aspx');
+        });
+        expect(ChoicesOverviewLinks.length).to.equal(constants.numberOfNearbyResultsToDisplay + 1);
```
st3v3nhunt: Good idea to use the constant to save having to change it when/if that value changes. Could the `1` also be replaced with the appropriate constant?

test/unit/lib/choicesOverview.js
```diff
@@ -0,0 +1,14 @@
...
+const chai = require('chai');
+
+const expect = chai.expect;
+
+describe('choicesOverview', () => {
+  it('should have valid ID appendend to the URL', () => {
+    const validId = [{
+      identifier: 'FA008'
+    }];
+    const fakeUrl = choicesOverview.addUrl(validId);
```
st3v3nhunt: The returned value isn't a fakeUrl, it's an array of cut down orgs. The variable name should reflect that, it is confusing otherwise.

test/unit/lib/choicesOverview.js
```diff
@@ -0,0 +1,14 @@
+const choicesOverview = require('../../../app/lib/choicesOverview');
+const chai = require('chai');
+
+const expect = chai.expect;
+
+describe('choicesOverview', () => {
+  it('should have valid ID appendend to the URL', () => {
```
st3v3nhunt: This test doesn't test each value in the input array is returned correctly. Please add another test for this scenario.

test/integration/resultsPagePlace.js
ajthomascouk: Done

test/unit/lib/choicesOverview.js
ajthomascouk: Done

test/unit/lib/choicesOverview.js
ajthomascouk: Done

st3v3nhunt: There is still no check for multiple values in the input list returning different values for `choicesOverviewUrl`. This leaves the implementation possibly being changed and the test not picking up on it.

ajthomascouk: Done

PR #399, opened 2017-11-24
display exact query search in results

markysoft:
## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 - Move 'none' option out of disambiguation place list
 - Move 'back' option out of list
 - Update npm dependencies
-- Display actual search term when no matches in results.
+- Display actual search term when no place or postcodes match
 
 0.20.0 / 2017-11-28
 ===================
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 - Move 'none' option out of disambiguation place list
 - Move 'back' option out of list
 - Update npm dependencies
+- Display actual search term when no matches in results.
 
 0.20.0 / 2017-11-28
 ===================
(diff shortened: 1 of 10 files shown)


app/views/find-help.nunjucks
```diff
@@ -4,7 +4,7 @@
 <h1 class="local-header--title--question">Find a pharmacy</h1>
 {% endblock %}
 
-{% block pageTitle %}Find a pharmacy{% if errorMessage %} - We can't find the postcode '{{location}}'{% endif %}{% endblock %}
+{% block pageTitle %}Find a pharmacy{% if errorMessage %} - We can't find the postcode '{{req_location}}'{% endif %}{% endblock %}
```
st3v3nhunt: Why is the title using the direct input and not the uppercased version as is used in the error message? I'd expect them to be the same.
I'd say having the title like this is correct but the message for inputs that are seen as postcodes but are invalid should also display the actual input rather than the uppercased input. What do you reckon?

CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+0.20.1 / TBA
+===================
+- Display actual search term when no matches in results.
```
st3v3nhunt: This confused me a bit as it isn't when there are no (pharmacy) results but when there are no place/postcode results. How about `Display actual search term when no place or postcodes match`?

app/views/find-help.nunjucks
markysoft: changed to show exact term entered.

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 - Move 'none' option out of disambiguation place list
 - Move 'back' option out of list
 - Update npm dependencies
+- Display actual search term when no matches in results.
```
st3v3nhunt: Any chance of this being updated. See previous https://github.com/nhsuk/connecting-to-services/pull/399#discussion_r153731532

PR #411, opened 2017-12-03
:christmas_tree: Display message if bank holiday coming up

st3v3nhunt:


app/lib/messages.js
```diff
@@ -10,8 +13,20 @@ function technicalProblems() {
...
 }
 
+function bankHolidayToday() {
+  return 'Today is a bank holiday. Please call to check opening times.';
+}
+
+function bankHolidayFuture(now, nextOpen) {
+  return isNextOpenTomorrow(now, nextOpen)
+    ? 'Tomorrow is a bank holiday. Please call to check opening times.'
+    : `${constants.dayOfWeekPrefixes[new Date(nextOpen).getDay()]}day is a bank holiday. Please call to check opening times.`;
```
markysoft: I'd probably put the code into a `getDayOfWeek(nextOpen)` function and move the word 'day' into the array of names to make it easier to read.

test/integration/bankHolidayAlert.js
```diff
@@ -0,0 +1,67 @@
...
+  afterEach('reset DATE', () => {
+    process.env.DATE = '2017-12-12';
+  });
+
+  describe('on the day of the bank holiday', () => {
+    before('set DATE', () => {
+      process.env.DATE = '2017-12-25';
+    });
+
+    it('should show a message about the bank holiday for each open result', (done) => {
```
markysoft: Does this mean that every pharmacy that is open in the list of results should have a message next to it, or only the one in the 'nearest open' section? 

From the implementation it looks like it is just the top section, as it's only looking for one message, which is the number of configured results in that section.

test/integration/bankHolidayAlert.js
```diff
@@ -0,0 +1,67 @@
...
+        .reply(200, serviceApiResponse);
+
+      chai.request(server)
+        .get(resultsRoute)
+        .query({ location: yourLocation, latitude, longitude })
+        .end((err, res) => {
+          iExpect.htmlWith200Status(err, res);
+          const $ = cheerio.load(res.text);
+
+          expect($('.callout--warning').first().text()).to.equal('Today is a bank holiday. Please call to check opening times.');
```
markysoft: I expect it's very tricky, but is it possible to write a test for tomorrow being a bank holiday?

app/lib/constants.js
```diff
@@ -9,4 +9,6 @@ module.exports = {
   placeSearch: 'place',
   postcodeSearch: 'postcode',
   yourLocationSearch: 'location',
+  dayInMilliseconds: 24 * 60 * 60 * 1000,
+  dayOfWeekPrefixes: ['Sun', 'Mon', 'Tues', 'Wednes', 'Thurs', 'Fri', 'Satur'],
```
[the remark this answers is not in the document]
st3v3nhunt: Fair enough, I don't think it  is odd - hence doing it this way.
Looking at the implementation again I can see the value of making it easier to read.

test/integration/bankHolidayAlert.js
st3v3nhunt: I can see the confusion I'll try to make it clearer. Every open org should get the message and in fact this is now extended to be every org that is open today.

test/integration/bankHolidayAlert.js
st3v3nhunt: I don't think the integration tests need to do more than check for the presence of the message. The different scenarios are (should be) covered by the unit tests for `addBankHolidayMessage`.

st3v3nhunt:
@markysoft I've taken on board the comments and made changes. As far as the integration test is concerned I've tried to make it clearer about each open result having a message but I've not extended it the bank holiday tomorrow as this is covered by the unit tests.

docker-compose-tests.yml
```diff
@@ -32,12 +32,12 @@ services:
       - nearby-services-api:nearby-services-api
 
   nearby-services-api:
-    image: "nhsuk/nearby-services-api:${NEARBY_SERVICES_API_DOCKER_IMAGE_TAG-0.5.2}"
+    image: "nhsuk/nearby-services-api:${NEARBY_SERVICES_API_DOCKER_IMAGE_TAG-0.13.0}"
     links:
-      - pharmacy-db:mongo
+      - pharmacy-db-elastic:es
```
markysoft: surprised we went so long with this still being called mongo

README.md
```diff
@@ -14,6 +14,14 @@ meet their time, location and accessibility needs.
 Clone the repo: `git clone https://github.com/nhsuk/connecting-to-services.git`
 and review the [`scripts`](scripts) to get up and running.
 
+## Changing the time
+
+The application will display a message for each open pharmacy on a bank holiday.
+In order to test this functionality the date can be set when the application
+starts. In order to do this an environment variable `DATE` needs to be set. For
+example, if the date wanted to be to set to Christmas day 2017 the following
+can be run on the command line `DATE=2017-12-25 ./scrtipts/start`.
```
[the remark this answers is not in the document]
st3v3nhunt: Good spot. Sorted.

PR #412, opened 2017-12-04
:nail_care: add url for blank personalisation links

ajthomascouk:


app/lib/fullUrlUtils.js
```diff
@@ -0,0 +1,15 @@
+function url(req) {
+  const protocol = req.protocol;
+  const host = req.get('host');
```
st3v3nhunt: Is there a reason this doesn't use `req.hostname`?
http://expressjs.com/en/4x/api.html#req.hostname

app/lib/fullUrlUtils.js
```diff
@@ -0,0 +1,15 @@
...
+  const originalUrl = req.originalUrl;
+  // eslint-disable-next-line no-script-url
+  return `${protocol}${host}${originalUrl}`;
```
st3v3nhunt: This needs to be include `://` i.e. `${protocol}://${host}${originalUrl}`.
It also looks like the protocol is coming through as `http` which will be because the SSL termination happens at the proxy. It might be worth hardcoding this to be `https://${host}${originalUrl}`. What do you reckon?

app/views/includes/header.nunjucks
```diff
@@ -41,8 +41,8 @@
         {% include "includes/header-items.nunjucks" %}
       </ul>
       <ul class="main-nav__account-menu account-menu">
-        <li><a href="#">Log in</a></li>
-        <li><a href="#">Create an account</a></li>
+        <li><a href="https://www.nhs.uk/Personalisation/Login.aspx">Log in</a></li>
```
st3v3nhunt: Why does the login page not have a return URL?

app/lib/fullUrlUtils.js
```diff
@@ -0,0 +1,15 @@
+function url(req) {
```
st3v3nhunt: This module doesn't need to be complex as this. I reckon you could just have the following:
```js
module.exports = req => `https://${req.hostname}${req.originalUrl}`;
}
```

I'd also rename it to be `completeOriginalUrl`.

app/middleware/setLocals.js
```diff
@@ -8,7 +9,11 @@ function fromRequest(req, res, next) {
     latitude: req.query.latitude,
     longitude: req.query.longitude,
   };
+  const fullUrl = fullUrlUtils(req);
   const backLink = backLinkUtils(req);
+  res.locals.fullUrl = {
```
st3v3nhunt: See comments about simplifying the returned value. If the suggestion is implemented this can be simplified to just `res.locals.completeOriginalUrl = completeOriginalUrl(req)`;

test/unit/lib/fullUrlUtils.js
```diff
@@ -0,0 +1,18 @@
+const chai = require('chai');
+const fullUrlUtils = require('../../../app/lib/fullUrlUtils');
+
+const expect = chai.expect;
+
+describe('fullUrlUtils', () => {
+  function host() { return 'beta.nhs.uk'; }
+  const mockReturnUrl = 'https://beta.nhs.uk/find-a-pharmacy/results?location=ls';
+
+  describe('url', () => {
```
st3v3nhunt: The test will need refactoring if the suggested changes are implemented.

app/views/includes/header.nunjucks
ajthomascouk: Pass. It doesn't on the main choices site.

test/unit/lib/fullUrlUtils.js
ajthomascouk: Done

app/lib/fullUrlUtils.js
ajthomascouk: Done

app/lib/fullUrlUtils.js
ajthomascouk: I read somewhere that you should use `req.get('host');`. However, I've tried your change and it works a treat.

app/lib/fullUrlUtils.js
ajthomascouk: Yup, seems sensible to me. Done

app/middleware/setLocals.js
ajthomascouk: Done

app/middleware/setLocals.js
```diff
@@ -9,11 +9,8 @@ function fromRequest(req, res, next) {
     latitude: req.query.latitude,
     longitude: req.query.longitude,
   };
-  const fullUrl = fullUrlUtils(req);
   const backLink = backLinkUtils(req);
-  res.locals.fullUrl = {
-    href: fullUrl.url,
-  };
+  res.locals.completeOriginalUrl = completeOriginalUrl(req);
```
st3v3nhunt: Can this line be moved above the creation of the const? It's a bit odd to have it in the middle of a variable declaration and the use of that variable.

app/views/includes/header.nunjucks
st3v3nhunt: Mmh, weird.

test/unit/lib/completeOriginalUrl.js
```diff
@@ -0,0 +1,13 @@
+const chai = require('chai');
+const completeOriginalUrl = require('../../../app/lib/completeOriginalUrl');
+
+const expect = chai.expect;
+
+describe('completeOriginalUrl', () => {
+  const reqMock = { hostname: 'beta.nhs.uk', originalUrl: '/find-a-pharmacy/results?location=ls' };
+  const mockOriginalUrl = 'https://beta.nhs.uk/find-a-pharmacy/results?location=ls';
```
st3v3nhunt: This isn't a mock, it is the expected output. Can the variable name be updated accordingly please?

test/unit/lib/completeOriginalUrl.js
```diff
@@ -0,0 +1,13 @@
...
+  it('should return the current URL', () => {
```
st3v3nhunt: Can this statement be expanded to incorporate the fact the returned URL will be `https` please?

ajthomascouk: Done

test/unit/lib/completeOriginalUrl.js
ajthomascouk: Done

app/middleware/setLocals.js
ajthomascouk: Done

PR #413, opened 2017-12-04
refactor headerItems

markysoft:
## Description
Improve code and and tests

## Checklist


test/unit/lib/buildHeaderItems.js
```diff
@@ -0,0 +1,84 @@
...
+      `<li><a href="https://www.nhs.uk/${partUrl}">${title}</a><ul>` +
+      `<li><a href="https://www.nhs.uk/${subMenu1PartUrl}">${subMenu1Title}</a></li>` +
+      `<li><a href="https://www.nhs.uk/${subMenu2PartUrl}">${subMenu2Title}</a></li>` +
+      '</ul></li>';
+    expect(output).to.equal(expectedHtml);
+  });
+
+  it('should return html for full headerItems response', () => {
+    const output = buildHeaderItems(headerApiResponse);
+    expect(output).to.exist;
```
st3v3nhunt: This seems a bit lazy. How about saving a copy of the generated file and comparing the output to that? This would act as a kind of integration test.

test/unit/lib/buildHeaderItems.js
```diff
@@ -0,0 +1,84 @@
...
+    expect(output).to.equal('');
+  });
+
+  it('should gracefully handle undefined urls', () => {
+    const output = cleanUrl(undefined);
+    expect(output).to.equal('');
+  });
+});
+
+describe('buildHeaderItems', () => {
```
st3v3nhunt: Can this be a in separate file to the execution of the tests for `cleanUrl`?

markysoft: will do, was in the same file as they are so closely related. Will also rename it 'cleanHeaderUrl to make it explicit that it's a url from the header api

package.json
```diff
@@ -21,10 +21,10 @@
     "postrewrite": "yarn git-hook",
     "precommit": "yarn git-hook",
     "prepush": "yarn git-hook && yarn snyk test",
-    "start": "yarn build-header && node app.js",
-    "start-watch": "yarn build-header && yarn brunch watch & nodemon app.js | ./node_modules/bunyan/bin/bunyan",
-    "test": "yarn build-header && NODE_ENV=test mocha --exit --recursive test",
-    "test-ci": "yarn build-header && yarn ci-hook && yarn coverage-upload-coveralls",
+    "start": "yarn header-build && node app.js",
```
st3v3nhunt: Does `header-build` need to be run here given it is running from the `Dockerfile` and going through the `Dockerfile` is the only way the application is going to start up correctly.

test/resources/headerItems.html
```diff
@@ -0,0 +1 @@
...s://www.nhs.uk/Service-Search/Pharmacy/LocationSearch/10">Pharmacies<span class="hidden"> search</span></a></li><li><a href="https://www.nhs.uk/Service-Search/Care%20at%20home/LocationSearch/1833">Homecare and care homes<span class="hidden"> search</span></a></li><li><a href="https://www.nhs.uk/Service-Search/Depression/LocationSearch/288">Depression<span class="hidden"> search</span></a></li><li><a href="https://www.nhs.uk/Service-Search/Maternity-service/LocationSearch/1802">Maternity<span class="hidden"> search</span></a></li><li><a href="https://www.nhs.uk/Service-Search/Alcohol-addiction/LocationSearch/1805">Alcohol<span class="hidden"> search</span></a></li><li><a href="https://www.nhs.uk/Service-Search/Sexual-health-services/LocationSearch/1847">Sexual health<span class="hidden"> search</span></a></li><li><a href="https://www.nhs.uk/Service-Search/Hospital/LocationSearch/7/Consultants">Consultants<span class="hidden"> search</span></a></li><li><a href="https://www.nhs.uk/Service-Search/Stop-smoking-services/LocationSearch/1846">Stop smoking services</a></li><li><a href="https://www.nhs.uk/Service-Search/Urgent-Care/LocationSearch/1824">Urgent care services</a></li></ul></li>
```
st3v3nhunt: Can this file end with a new line char or does that mess up the test? I assume this is the generated content? If its a problem don't worry about it.

docker-compose.yml
```diff
@@ -8,12 +8,8 @@ services:
       args:
         NODE_ENV: development
     command: yarn run start-watch
-    volumes:
-      - .:/code
```
st3v3nhunt: Why has this been removed? I think the result of not having this here is changes made to the code are not going to be picked up in the `test-watch` and `start-watch` scripts when run in the Docker container.

CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+0.21.1 / 2017-12-12
```
st3v3nhunt: The change feels a bit bigger than a patch version. I would have made it a new minor version but that is just my opinion.

docker-compose.yml
markysoft: mounting the source code folder replaces the code folder in the docker image, which breaks the CI tests build as there is no header-items.nunjucks in the source folder.

I'll add the volume specifically to the start and test sections to ensure changes are seen. These include the header-build as part of the task, to ensure the file is present.

package.json
markysoft: without the header-build in the task this won't work locally if the 'code' folder is mounted.
Start is not used in the container, the app is started directly to avoid multiple thread issues.

st3v3nhunt: Good point about how the container starts.
Is the `start` script used for anything? Could it be removed?

docker-compose.yml
st3v3nhunt: I can see the point of removing the mapping in `docker-compose-tests` but in this file I think it is required. Or this needs moving from the `base-app` service to the `frontend` service.

test/resources/headerItems.html
markysoft: Adding a new line would mean also adding or removing a new line in the test assert, so best left as it is really.

package.json
markysoft: It can go, it's not used anywhere. I'll remove it.

PR #419, opened 2017-12-13
Feature/more results extend

st3v3nhunt:


CHANGELOG.md
```diff
@@ -1,7 +1,8 @@
-0.22.0 / 2017-12-12
+0.23.0 / 2017-12-14
```
ajthomascouk: Why are we skipping `0.22.0` ?

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "connecting-to-services",
-  "version": "0.22.0",
+  "version": "0.23.0",
```
ajthomascouk: Same question as above.

CHANGELOG.md
ajthomascouk: why are we skipping `0.22.0` ?

package.json
ajthomascouk: Same question as above.

CHANGELOG.md
st3v3nhunt: It has already been created.

ajthomascouk: aah `0.22.0` was `Pre-release` I thought it was just `Draft`. All good then.

PR #434, opened 2018-01-17
:nail_care: activate toggle

ajthomascouk:


app/public/js/init.js
```diff
@@ -2,3 +2,4 @@
 require('./cookieMessage');
 require('./geoLocation');
 require('./menuToggle');
+require('./c2sScripts');
```
st3v3nhunt: Can this script be renamed to be less generic? Something like `toggleResults` would make it easier to know what is does.

app/public/js/c2sScripts.js
```diff
@@ -0,0 +1,9 @@
+(function (global) {
+  'use strict';
+  $('.viewToggle a').not('.checked').attr('href', function() {
+    return this.href + '&open=true';
+  });
+  $('.viewToggle a.checked').attr('href', function() {
+    return this.href = window.location.href.replace(/&open=[^&;]*/,'');
```
st3v3nhunt: Use a local copy of `window` as in https://github.com/nhsuk/connecting-to-services/blob/master/app/public/js/geoLocation.js#L9

app/public/js/c2sScripts.js
```diff
@@ -0,0 +1,9 @@
+(function (global) {
+  'use strict';
+  $('.viewToggle a').not('.checked').attr('href', function() {
```
st3v3nhunt: Use a local copy of jQuery as in https://github.com/nhsuk/connecting-to-services/blob/master/app/public/js/geoLocation.js#L3

ajthomascouk:
> Thinking about this a bit more. There is no need to populate the href via JS. It could be done server side which has the benefit of not needing an additional script to be loaded, it would be quicker and it would work even if there is no JS in the client.

Make it so Captain

PR #435, opened 2018-01-17
:watch: Display real opening times

st3v3nhunt:


app/views/includes/openingTimes.nunjucks
```diff
@@ -0,0 +1,16 @@
+<table class="openingTimes">
+  <tr><th>Day of the week</th><th>First opening time</th><th>Last closing time</th></tr>
+  {% for openingTime in service.openingTimesOverview %}
+  <tr><th>{{ openingTime.day }}</th>
+    {% if openingTime.openingTimes === null %}
+    <td colspan="2">Unavailable</td>
+    {% elif openingTime.openingTimes.opens === null%}
```
st3v3nhunt: Assumption that if there is always both an `opens` and a `closes` so  only one needs to be checked in order to determine if it is closed.

PR #451, opened 2018-01-23
:clap: new layout

ajthomascouk:


app/views/includes/openingTimes.nunjucks
```diff
@@ -1,20 +1,31 @@
...
+          <tr>
+          {% if loop.length !== 1 %}
+            <th rowspan="{{loop.length}}">{{ openingTime.day }}</th>
+          {% else %}
+            <th>{{ openingTime.day }}</th>
+          {% endif %}
+            <td>{{sessions.opens}} to {{ "midnight" if sessions.closes === "11:59pm" else sessions.closes }}</td>
+          </tr>
+        {% else %}
+          <tr class="{{ "hasSessions" if loop.length !== 1 }}">
```
st3v3nhunt: This can be simplified to `<tr class="hasSessions">` as this branch will only run when the loop length is greater than 1.

app/views/includes/openingTimes.nunjucks
```diff
@@ -1,20 +1,31 @@
...
+      <tr>
+        <th>{{ openingTime.day }}</th>
+        <td>Closed</td>
+      </tr>
     {% endif %}
-  </tr>
   {% endfor %}
 </table>
-<p>May close during these hours</p>
-<p class="overviewHours"><a href="{{ service.choicesOverviewUrl }}">See exact hours</a></p>
```
st3v3nhunt: `choicesOverview` can be removed not it is no longer being used.

scss-c2s/c2s-results.scss
```diff
@@ -166,55 +192,75 @@ body {
                 }
               }
             }
+            &.wrapped {
```
st3v3nhunt: If this class isn't being added by the JS can it be removed?

PR #460, opened 2018-02-06
:kissing_heart: no times and wrapped style changes

ajthomascouk:


app/public/js/isWrapped.js
```diff
@@ -0,0 +1,11 @@
+(function (global) {
+  'use strict';
+  var $ = global.jQuery;
+  $('h2.results__name').each(function(){
+    var $height = $(this).height();
```
st3v3nhunt: The convention to prepend variable names with `$` is only applicable when that variable is a jQuery element. It can be removed in this instance as height is just a number.

app/public/js/isWrapped.js
```diff
@@ -0,0 +1,11 @@
...
+    var $lineHeight = Math.floor(parseInt($(this).css('line-height').replace('px','')));
```
st3v3nhunt: Same as above, the variable can just be `lineHeight`.
Can you add the radix to the parseInt call please e.g.
`var lineHeight = Math.floor(parseInt($(this).css('line-height').replace('px',''), 10));`
See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/parseInt for more info.

app/public/js/noTimes.js
```diff
@@ -0,0 +1,7 @@
+(function (global) {
+  'use strict';
+  var $ = global.jQuery;
+  $('p.callout--info__mesg:contains("We can\'t find any opening times")').each(function(){
```
st3v3nhunt: Mmh, this isn't great. Although I can't find a better way to do it. Can the string be a little less strict, something like just 'find any opening times'? The message comes from the API so it is going to be difficult to detect changes.
@markysoft any ideas?

markysoft: I think that message occurs when there are no opening times on the services object, so you could check if the service.openingTimes is undefined (or closed all week?) and add a 'noOpeningTimes' property to the service

st3v3nhunt: In fact, I don't think this needs to be done on the client at all.

PR #461, opened 2018-02-07
Feature/moar tests

st3v3nhunt:


test/integration/bankHolidayAlert.js
```diff
@@ -49,20 +49,19 @@ describe('The bank holiday alert messaging', () => {
         .times(1)
         .reply(200, serviceApiResponse);
 
-      chai.request(server)
+      return chai.request(server)
```
neilbmclaughlin: The use of promises in the tests looks good. Some of the chai.request() statements are returned and others not and I can't figure out why. Could you elaborate please.

st3v3nhunt: They are all returned, just some are done implicitly.
https://eslint.org/docs/rules/arrow-body-style

st3v3nhunt: Returning the promise is required https://mochajs.org/#working-with-promises

test/integration/app.js
```diff
@@ -10,25 +10,24 @@ const expect = chai.expect;
...
-        iExpect.htmlWith200Status(err, res);
+      .then((res) => {
+        iExpect.htmlWith200Status(res);
 
         expect(res).to.redirect;
         expect(res.req.path).to.equal(`${constants.SITE_ROOT}/`);
-        done();
-      });
+      })
+      .catch((err) => { throw err; });
```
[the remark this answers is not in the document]
st3v3nhunt: I think I'd prefer to use promises rather than async, await.

test/integration/bankHolidayAlert.js
markysoft: I see, so it's the lack of the return of the 'chai.request()' promise that is causing the silent failure.
In that case, all tests should return the chai.request(), and the catch-throw isn't necessary so can be removed from the tests.

st3v3nhunt: Not quite as `chai.request` is being returned implicitly. I'll finish the changes to use async, it looks quite a bit better.

markysoft: That's the theory, but the 'unknown page' test below fails silently with no return from chai.request, but correctly reports the test as failing if you **do** return the value.

test/integration/app.js
```diff
@@ -1,42 +1,36 @@
...
-        const $ = cheerio.load(res.text);
-
-        expect($('.local-header--title--question').text().trim())
-          .to.equal('Page not found');
-        expect($('title').text()).to.equal('Page not found - NHS.UK');
-      })
-      .catch((err) => { throw err; });
+  it('should return a 404', async () => {
+    try {
+      await chai.request(server).get(`${constants.SITE_ROOT}/not-known`);
```
neilbmclaughlin: Should there be an `expect.fail()` here for if no exception is thrown?

neilbmclaughlin: Or, possibly, use expect.throw(..)

PR #463, opened 2018-02-12
:earth_africa: prevent start address from appearing for place searches

markysoft:
## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 ===================
 - Additional tests for opening times
 - Update npm dependencies
+- Prevent start address from appearing in map link for place searches
 
 0.30.0 / 2018-02-06
 ===================
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-0.31.0 / 2018-02-08
+0.31.0 / 2018-02-15
 ===================
 - Additional tests for opening times
 - Update npm dependencies
--- a/test/unit/lib/getSearchType.js
+++ b/test/unit/lib/getSearchType.js
@@ -1,0 +1,35 @@
+const chai = require('chai');
+const getSearchType = require('../../../app/lib/getSearchType');
+const constants = require('../../../app/lib/constants');
+
+const expect = chai.expect;
+
+describe('get search type', () => {
+  it('should return postcodeSearch when location is a postcode', () => {
+    const noCoordinates = {};
+    const type = getSearchType('LS1', noCoordinates);
+
+    expect(type).to.equal(constants.postcodeSearch);
+  });
+
+  it('should return placeSearch when location is not a postcode', () => {
+    const noCoordinates = {};
+    const type = getSearchType('Leeds', noCoordinates);
+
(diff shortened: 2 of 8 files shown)


test/unit/lib/mapLink.js
```diff
@@ -85,6 +85,57 @@ describe('mapLink', () => {
...
+      };
+      const expectedMapLink = `https://maps.google.com/maps?${qs.stringify(params)}`;
+
+      const results = mapLink.addUrl(yourLocationSearchCriteria, inputList);
+
+      expect(results.length).to.be.equal(1);
+      expect(results[0].mapUrl).to.be.equal(expectedMapLink);
+    });
+
+    it('should populate start address for place searches', () => {
```
st3v3nhunt: The search in the test is a postcode rather than a place (as stated in the test name).

test/unit/lib/mapLink.js
```diff
@@ -85,6 +85,57 @@ describe('mapLink', () => {
...
+      const inputList = [{
+        name: 'name',
+        address: {
+          line1: 'line1',
+        },
+      }];
+      const destination = 'name,line1';
+      const params = {
+        saddr: '',
+        daddr: `${destination}`,
```
st3v3nhunt: Could this just be `daddr: destination` rather than interpolating the variable?
Same goes for the following line and the two reference in the test below.

test/unit/lib/getSearchType.js
```diff
@@ -0,0 +1,35 @@
...
+  });
+
+  it('should return placeSearch when location contains postcode and coordinates, i.e. from the disambiguation page', () => {
+    const coordinates = { latitude: 0.1, longitude: -0.5 };
+    const type = getSearchType('Leeds LS1', coordinates);
+
+    expect(type).to.equal(constants.placeSearch);
+  });
+
+  it('should return yourLocationSearch when location is \'your Location\'', () => {
```
st3v3nhunt: As the coordinates play no role in the determination of the search type as `yourLocation` can another test be added where there are no coordinates to make that clear?

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 ===================
```
st3v3nhunt: Seen as this date is no in the past could you update it to a possible future release date please?

test/unit/lib/mapLink.js
```diff
@@ -85,6 +85,57 @@ describe('mapLink', () => {
       expect(results[0].mapUrl).to.be.equal(expectedMapLink);
     });
 
+    it('should leave start address empty for place searches', () => {
```
Dan-J-Scho: Just slightly unclear as to the difference between this unit test and the one which has also been added below below without digging into the code.

test/unit/lib/mapLink.js
markysoft: Didn't register that it had been written like that, I've corrected them now.

test/unit/lib/mapLink.js
markysoft: Thanks, the other one should have said postcode, so should be clearer now.

PR #467, opened 2018-02-20
:eyes: Fix incorrect reference to map link

st3v3nhunt:


test/integration/resultsPagePostcode.js
```diff
@@ -51,7 +51,7 @@ describe('The results page', () => {
     const results = $('.results__item');
     expect(results.length).to.equal(nearbyResultsCount);
 
-    const mapLinks = $('.results__maplink a');
+    const mapLinks = $('.maplink');
     mapLinks.toArray().forEach((link) => {
       expect($(link).attr('href')).to.have.string(`https://maps.google.com/maps?saddr=${encodeURIComponent(ls27ue)}`);
     });
```
markysoft: Could you add an assert for the mapLinks array length to be greater than zero so it can't fail silently?

test/integration/resultsPagePostcode.js
```diff
@@ -88,7 +88,7 @@ describe('The results page', () => {
     const results = $('.results__item');
     expect(results.length).to.equal(numberOfResults);
 
-    const mapLinks = $('.results__maplink a');
+    const mapLinks = $('.maplink');
     mapLinks.toArray().forEach((link) => {
```
markysoft: See above comment

st3v3nhunt: Changes made and another test that wasn't doing anything fixed.

PR #470, opened 2018-02-21
Optimisation of  Dockerfile to minimise layers 

neilbmclaughlin:
@st3v3nhunt - in order to understand docker a bit more I've tried to identify some improvements and note the rational for the changes. I'd appreciate your comments. Also, there should be no functional differences so feel free to reject the PR.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -23,10 +23,13 @@ ENV NODE_ENV=${NODE_ENV}
 
 WORKDIR /code
 
+COPY --chown=nodeuser:nodeuser yarn.lock package.json /code/ 
+
+RUN if [ "$NODE_ENV" == "production" ]; then yarn install --production --pure-lockfile; else yarn install --pure-lockfile; fi
+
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -1,31 +1,36 @@
...
-COPY yarn.lock package.json /code/
-
-RUN if [ "$NODE_ENV" == "production" ]; then yarn install --production --pure-lockfile; else yarn install --pure-lockfile; fi
 
-EXPOSE 3000
+WORKDIR /code
 
-COPY . /code
+# .dockerignore prevents unneeded files from being copied
+COPY --chown=nodeuser:nodeuser . .
```
neilbmclaughlin: On reflection I think this COPY should be split as it was before - one COPY for the package info and a subsequent one for the code.

st3v3nhunt:
Closing. If necessary it can be picked up at a later date.

PR #483, opened 2018-03-21
Feature/switch results

st3v3nhunt:
Show open results by default when request is made at point deemed out of business hours i.e. at the weekend, before 08:00, after 18:00 or on a bank holiday.

Refactoring around dateUtils and a bunch of unit tests are also included.


README.md
```diff
@@ -69,7 +70,7 @@ environment.
...
 | `BASIC_AUTH`                       | An MD5 encrypted [htpasswd](https://httpd.apache.org/docs/2.4/misc/password_encryptions.html)                                                              | test:test                 |            |
 | `DISABLE_GOOGLE_SEARCH`            | Set to 'true' to disable the metadata which enables [Google Sitelinks searchbox](https://developers.google.com/search/docs/data-types/sitelinks-searchbox) | false                     |            |
-| `DATE`                             | Set the date used to determine if a bank holiday message should be displayed                                                                               |                           | No         |
+| `DATETIME`                         | Set the datetime used to determine if a bank holiday message should be displayed or the behaviour of the results page                                                                  |                           | No         |
```
markysoft: The description isn't entirely clear upon first reading, but I'm struggling to think of a better way to word it. 
There's more spaces on the line than other env vars so the table fields don't line up.

app/lib/constants.js
```diff
@@ -9,6 +9,16 @@ module.exports = {
...
+  businessHours: {
+    end: {
+      hour: 18,
+      minute: 0,
+    },
+    start: {
+      hour: 8,
+      minute: 0,
+    },
+  },
```
markysoft: Should the business hours/minutes be exposed as env vars so they can be easily changed without a code change?

README.md
st3v3nhunt: I've given it a go i.e. trying to clarify the use of the env var.

app/lib/constants.js
st3v3nhunt: I did consider this but didn't do. I've added it in now.

PR #497, opened 2018-04-30
Use first non-empty address line in service/map screen reader link

markysoft:
## Description
Use first non-empty address line in screen reader only text for Map and Direction and Services Available links

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 1.2.0 / 2018-05-3
 ===================
 - Use first non-empty address line in screen reader only text for Map and Direction and Services Available links
+- Allow fonts from hotjar.com in Content Security Policy
 
 1.1.0 / 2018-04-17
 ===================
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+1.2.0 / 2018-05-3
+===================
+- Use first non-empty address line in screen reader only text for Map and Direction and Services Available links
+
 1.1.0 / 2018-04-17
 ===================
 - Update npm dependencies
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-1.2.0 / 2018-05-3
+1.2.0 / 2018-05-03
 ===================
 - Use first non-empty address line in screen reader only text for Map and Direction and Services Available links
 - Allow fonts from hotjar.com in Content Security Policy
--- a/app/middleware/prerender.js
+++ b/app/middleware/prerender.js
@@ -1,7 +1,21 @@
+const addAddressLine = require('../lib/addAddressLine');
 const addBankHolidayMessage = require('../lib/addBankHolidayMessage');
 const choicesServices = require('../lib/choicesServices');
 const mapLink = require('../lib/mapLink');
 const updateOpeningTimes = require('../lib/updateOpeningTimes');
+
+function updateServices(searchCriteria, services) {
+  if (services && Array.isArray(services)) {
+    /* eslint-disable no-param-reassign */
+    services = mapLink.addUrl(searchCriteria, services);
+    services = choicesServices.addUrl(services);
+    services = addBankHolidayMessage(services);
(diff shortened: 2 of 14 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+1.2.0 / 2018-05-3
```
st3v3nhunt: Can the day format be dd e.g. `03`, thanks.

app/middleware/prerender.js
```diff
@@ -1,19 +1,30 @@
+const addAddressLine = require('../lib/addAddressLine');
 const addBankHolidayMessage = require('../lib/addBankHolidayMessage');
 const choicesServices = require('../lib/choicesServices');
 const mapLink = require('../lib/mapLink');
 const updateOpeningTimes = require('../lib/updateOpeningTimes');
 
+function updateServices(searchCriteria, services) {
+  if (services && Array.isArray(services)) {
+    /* eslint-disable no-param-reassign */
```
st3v3nhunt: It would be nice to not be manipulating the input in all these functions. Something for a future refactor.

config/express.js
```diff
@@ -29,51 +30,7 @@ module.exports = (app, config) => {
...
-          '\'self\'',
-          '\'unsafe-inline\'',
-          '*.nhs.uk',
-        ],
-      },
-    },
-    frameguard: { action: 'deny' },
-    hsts: { includeSubDomains: false },
-  }));
+  helmet(app);
```
st3v3nhunt: This approach makes it much cleaner.

test/unit/lib/addAddressLine.js
```diff
@@ -0,0 +1,118 @@
...
+  {
+    address,
+    name: nameTwo,
+  }];
+
+  return services;
+}
+
+describe('addAddressLine', () => {
+  it(
```
st3v3nhunt: All of the text descriptions for the `it` blocks start on the following line but the lines don't look long enough to fail due to length. Is there another reason why they are formatted this way? If not, can they be brought back to a single line?

test/unit/lib/addAddressLine.js
```diff
@@ -0,0 +1,118 @@
...
+        delete s.address;
+      });
+
+      const results = addAddressLine(services);
+
+      expect(results).to.not.be.undefined;
+      expect(results).to.be.an('array');
+      expect(results.length).to.be.equal(2);
+      expect(results[0].address).to.be.undefined;
+      expect(results[1].addressLine).to.be.undefined;
```
st3v3nhunt: Should these last 2 expectations be duplicated but for the other result item? e.g.
```js
expect(results[0].address).to.be.undefined;
expect(results[0].addressLine).to.be.undefined;
expect(results[1].address).to.be.undefined;
expect(results[1].addressLine).to.be.undefined;
```

test/unit/lib/addAddressLine.js
```diff
@@ -0,0 +1,118 @@
...
+        delete s.address.line2;
+        delete s.address.line3;
+      });
+
+      const results = addAddressLine(services);
+
+      expect(results).to.not.be.undefined;
+      expect(results).to.be.an('array');
+      expect(results.length).to.be.equal(2);
+      expect(results[1].addressLine).to.be.undefined;
```
st3v3nhunt: Similar to the comment above, should or could there be another expectation for the other item i.e. `expect(results[0].addressLine).to.be.undefined;`

app/middleware/prerender.js
markysoft: Agreed, it's a bit odd modifying the parameter, and returning the value, as it's not obvious there are side affects.

test/unit/lib/addAddressLine.js
markysoft: Overzealous code formatter, have improved it now, thanks.

app/middleware/prerender.js
st3v3nhunt: Indeed. It does make it more fun trying to reason about the code when you need to have half a dozen different files open!

PR #500, opened 2018-05-23
Update Webtrends and GA configuration for GDPR compliance

markysoft:
## Checklist


app/views/includes/analytics.nunjucks
```diff
@@ -1,46 +1,79 @@
...
+    <img alt="dcsimg" id="dcsimg" width="1" height="1" src="//statse.webtrendslive.com/{{ WEBTRENDS_ANALYTICS_TRACKING_ID }}/njs.gif?dcsuri=/nojavascript&amp;WT.js=No&amp;WT.tv=10.4.23&amp;dcssip=.nhs.uk&amp;dcsipa=1"/>
   </noscript>
 {% endif %}
+
+<script>
+  function CookiebotCallback_OnAccept() {
+    if (Cookiebot.consent.statistics) { 
+      enableGoogleAnalyticsUniversal(); 
+      enableWebtrends();
+      enableHotjar();
```
beseven: Hotjar is a `preference` consent rather than statistics - I also suggest splitting this in two partials - one that goes in `head` (Hotjar + Cookiebot can also live in that one) and one that lives in `body` (webtrends + GA)

markysoft: Thanks for the info, only a single `CookiebotCallback_OnAccept` can be defined though, so we have to stick with one partial.
I'll move the Hotjar logic to to a `preferences` consent check.

beseven: I guess my questions is if hotjar will work if not in head.

markysoft: Hotjar is put into head to 'work as soon as possible' https://help.hotjar.com/hc/en-us/articles/115009336727-Manually-Adding-the-Tracking-Code. 
In the new implementation we have to wait for consent before Hotjar is enabled so it needs to be loaded dynamically. The downside is we may miss a bit of tracking while the code fires, but I don't see any way around that.

beseven:
LGTM, tested by Shaun.

PR #527, opened 2018-11-09
Add datalayer

st3v3nhunt:
Fixes #524


app/views/includes/digitalData.nunjucks
```diff
@@ -0,0 +1,3 @@
+<script>
+  window.digitalData = {{ digitalData | dump | safe }};
```
neilbmclaughlin: Did you intend for `dump` to be in the commit? The notes[1] on `dump` suggest it's a debugging tool.

[1] https://mozilla.github.io/nunjucks/templating.html#dump

st3v3nhunt: Yeah, it is intentional. `digitalData` is just a JSON object so in order for it to be correct when 'dumped' into the view it needs to run through `JSON.stringify`.

PR #557, opened 2019-07-10
Feature/meta data

st3v3nhunt:


app/lib/constants.js
```diff
@@ -9,6 +9,17 @@ module.exports = {
       open: 'open',
     },
   },
+  app: {
```
beseven: nice use of constants

config/contentSecurityPolicy.js
```diff
@@ -5,6 +5,8 @@ module.exports = {
       'assets.adobedtm.com',
       '*.demdex.net',
       '*.hotjar.com:*',
+      '*.hotjar.io',
```
beseven: can you add this change to the changelog too please?

st3v3nhunt: Done

st3v3nhunt:
I've updated the dependencies (again) too.

st3v3nhunt:
Resolves #559

PR #561, opened 2019-07-17
Feature/full image path

st3v3nhunt:


app/lib/constants.js
```diff
@@ -20,12 +18,14 @@ module.exports = {
       site: '@nhsuk',
     },
   },
+  assetsUrl: 'https://assets.nhs.uk',
```
beseven: Is this a good thing to hard code? the assets in staging have different url

test/integration/metaData.js
```diff
@@ -13,21 +13,22 @@ const expect = chai.expect;
 
 chai.use(chaiHttp);
 
-const resultsRoute = `${constants.SITE_ROOT}/results`;
+const siteRoot = constants.siteRoot;
+const resultsRoute = `${siteRoot}/results`;
 
 function expectStandardMetadata($) {
-  const canonicalUrl = `https://127.0.0.1${constants.SITE_ROOT}/`;
+  const canonicalUrl = `https://127.0.0.1${siteRoot}/`;
```
beseven: should the cannonicalUrl have `/` at the end or not?

CHANGELOG.md
```diff
@@ -1,4 +1,10 @@
-1.13.0 / TBC
+1.14.0 / TBC
+===================
+- Use full path for Open Graph image
+- Do not use capitalised convention for constant variables
```
beseven: Any reasoning for this? ALL CAPS gives a visual indication that that variable should be set somewhere

app/lib/constants.js
st3v3nhunt: Possibly not. It might be better to use an env var, however, it hasn't changed in this PR. And if it did changed based on the env var and that was different in the pre-live environment you have a situation where the first time the actual assets are being tested is when the site is running in live. Which isn't the first place I want to know if something is working!

CHANGELOG.md
st3v3nhunt: Sure, however, this is more of a hangover from previous versions of JS that didn't have the `const` keyword.

PR #572, opened 2019-08-21
Feature/azure search tweaks

st3v3nhunt:
Possible tweaks for consideration to be merged into #567 
Changes include:
* `getContacts` is more flexible - if different types of contact are added (to the API) they will be returned by in the contacts object, keyed by the type. There is no longer any mapping between keys i.e. the view uses the same key as the type of the contact e.g. `telephone` no longer needs mapping to `telephoneNumber`. Performance should also improve (although it is likely to be just a negligible amount due to the size of the data). The array of contacts is no longer iterated for every type of contact, it is now iterated just once.


test/resources/organisations/FK276-es.json
```diff
@@ -129,15 +129,14 @@
     "postcode": "LS1 7JH"
   },
   "contacts": {
-    "website": "",
+    "website": "some.thing.com",
     "additionalContacts": [],
-    "telephoneNumber": "0113 245 5097",
+    "telephone": "0113 245 5097",
```
neilbmclaughlin: Just a comment here for future reference:

Although the format of the ES response is our starting point, this change in the renaming of the property is an example of the code diverging from that baseline format for clarity purposes (and this is a good thing).

PR #581, opened 2019-09-17
Reorganise tests using subdirectories

neilbmclaughlin:
Place mocha and backstop tests in their own subdirectores principally to allow selective test running.

Also took the opportunity to switch from istanbul to nyc for code coverage as istanbul is deprecated.

Note 1: The vast majority of the changed files are simply adding another parent to the relative path in `require`s of tests to the subject under test. These are automatically verified by eslint which will find and flag any unresolvable paths.


.gitignore
```diff
@@ -30,3 +30,6 @@ newrelic_agent.log
 backstop-tests/backstop_data/html_report/
```
st3v3nhunt: `test/` needs appending to this line and the two below.

neilbmclaughlin:
Changes made and (hopefully) ready for merge.

neilbmclaughlin:
I agree about making the test subdirectories technology agnostic and the above suggestion for a structure seems to work well. I'll scrap this PR and create a new one for this work.

PR #583, opened 2019-10-01
Greenkeeper/eslint config nhsuk 0.21.1

neilbmclaughlin:


neilbmclaughlin:
Did this in separate PR since the eslint upgrade introduced failing eslint checks

PR #586, opened 2019-10-02
Remove unneeded checks on OpeningTimeType from query clauses

neilbmclaughlin:


neilbmclaughlin:
Following discussion with @st3v3nhunt 

> I'm wondering if it is worth keeping the filter in. e.g. what happens if pharmacies start to have some other type of opening time e.g. an opening time for pickup only? Without the filter, this would be included?

> Better to have a whitelist than allow anything? There could be unintentional changes to the application based on the data.

PR #587, opened 2019-10-03
Guard against data inconsistency

neilbmclaughlin:
Additional opening times for phamacies (and possibly other org types)
can have a type of _either_ Additional or General.  This is an error in
the source data which seems to have been introduced in 2013 and never
corrected.  We have extended the guard rather than remove it (since
'additionalness' can be infered by a non-zero AdditionalOpeningDate) to
filter out any future opening times for different types (e.g. reception)

### DIFF
--- a/app/lib/queryBuilder.js
+++ b/app/lib/queryBuilder.js
@@ -1,4 +1,3 @@
-const moment = require('moment');
 const VError = require('verror').VError;
 const queryTypes = require('./constants').queryTypes;
 const getDateTime = require('./dateUtils').getDateTime;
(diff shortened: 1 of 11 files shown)


app/lib/queryBuilder.js
```diff
@@ -15,21 +15,24 @@ function getOpenPharmacyFilter(date) {
...
+  // This is an error in the source data which seems to have been introduced in 2013
+  // and never corrected.
+  // I think that additional hours should be of type general to allow, for example,
+  // a GP surgery to have different hours on a bank holiday for reception only.
   return `
   ${pharmacyFilter} and
   ( OpeningTimesV2/any(time:
-          time/IsOpen
           and time/Weekday eq '${weekday}'
           and time/OpeningTimeType eq 'General'
```
st3v3nhunt: I presume this doesn't need updating to include a `Additional OpeningTimeType` clause because the data is such that these times are only ever `General`?

neilbmclaughlin: That's correct.

neilbmclaughlin:
I've added a suite of tests for the queryBuilder to ensure that token substitution works and different queryTypes are handled correctly. I think the tests do add value.

neilbmclaughlin:
More timezone fun and games. Marked this as WIP until I've resolved it.

neilbmclaughlin:
Resolved the timezone issue and also added a guard for the scenario where a pharmacy has no contact details which was causing null ref exceptions during testing.

test/unit/lib/mappers/azContactsMapper.js
```diff
@@ -0,0 +1,92 @@
+
```
st3v3nhunt: Nit: No empty line at the beginning of the file.

PR #590, opened 2019-10-08
:question: Changes required for prefer-destructuring rule change

st3v3nhunt:


st3v3nhunt:
Closing PR as it was only meant as a way to review the likely changes with the update to the `prefer-destructuring` eslint rule.

PR #596, opened 2019-10-10
Add subtype to pharmacy filter

neilbmclaughlin:
This is to ensure only community pharmacies appear in results and
appliance/internet/etc do not


neilbmclaughlin:
Hold of merging, I need to update the change log and version.

st3v3nhunt:
Not sure why the `coveralls` check hasn't succeeded as the code coverage is fine.

PR #598, opened 2019-10-16
:flashlight: add useful diagnostics to metadata

neilbmclaughlin:


neilbmclaughlin:
I'll revisit this.
Is there any other useful diagnostics you can think of which we should add?

st3v3nhunt:
I can't think of anything off the top of my head that isn't already available from `/metrics` (which isn't available in prod and is meant for collection and analysis in Grafana) that I'd want to add to it.

PR #600, opened 2019-10-16
Add diagnostics page

neilbmclaughlin:


app/middleware/debugRenderer.js
```diff
@@ -0,0 +1,9 @@
+const { getDateTime } = require('../lib/dateUtils');
+const { version } = require('../../package');
+
+module.exports = (req, res) => {
+  res.json({
+    serverDateTime: getDateTime(),
```
st3v3nhunt: nit: Is it actually a server? 😝

config/routes.js
```diff
@@ -7,6 +7,7 @@ const locationValidator = require('../app/middleware/locationValidator');
 const logZeroResults = require('../app/middleware/logZeroResults');
 const prerender = require('../app/middleware/prerender');
 const renderer = require('../app/middleware/renderer');
+const debugRenderer = require('../app/middleware/debugRenderer');
```
st3v3nhunt: nit: I'm not sure if `Renderer` is required, it could quite easily just be `debug`. And then I think about whether `debug` is a good name. Perhaps `info` would be better?

config/routes.js
```diff
@@ -27,4 +28,9 @@ router.get(
   renderer.results
 );
 
+router.get(
+  '/debug',
```
st3v3nhunt: Should there be a test for the route given it might become a relied upon feature?

doc/0015-add-info-page.md
```diff
@@ -0,0 +1,19 @@
...
+
+Date: 2019-10-16
+
+## Status
+
+Accepted
+
+## Context
+
+When debugging it is useful to be sure of what the deployed version of the application is. In addition, when debugging date and time issues (e.g openness) then it is usefull to know the time in UTC from the point of view of the application.
```
st3v3nhunt: nit: typo on the second useful (double ll).

test/integration/infoPage.js
```diff
@@ -0,0 +1,19 @@
...
+
+chai.use(chaiHttp);
+
+describe('Info page', () => {
+  it('should display info', async () => {
+    const res = await chai.request(server).get(`${constants.siteRoot}/info`);
+    const jsonResponse = JSON.parse(res.text);
+    expect(jsonResponse).to.not.be.null;
+    expect(Object.keys(jsonResponse)).to.have.members(['version', 'now']);
+    expect(jsonResponse.version).to.not.be.null;
```
st3v3nhunt: This could be checked against the actual version in `package.json`.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "connecting-to-services",
-  "version": "1.15.5",
+  "version": "1.16.0",
```
st3v3nhunt: Should this be `1.17.0`?

neilbmclaughlin: Good spot, fixed.

PR #603, opened 2019-10-20
Fake time experiment

neilbmclaughlin:


st3v3nhunt:
Was used for testing, no longer required.

PR #606, opened 2019-10-23
Change the analytics base data layer ahead of move from beta to live

neilbmclaughlin:
Note: The term 'data layer' comes from how the analytics team would refer to the 'nhs:web' value


neilbmclaughlin:
Closing and recreating to trigger rebuild on Travis

Issue #609, opened 2019-10-24
Integration tests fail if run outside of business hours


neilbmclaughlin:
The behaviour of the app changes depending on the time of day, specifically in or out of business hours. The behaviour change is which view is presented for the results page - _nearby_ or _nearby + open_. During business hours the default view is _nearby_ and outside business hours the default view is _nearby + open_.The tests fail because the nock's are set up with an assumption of the in business hours default view and therefore what the form of the query to azure search will be to mock.

The likely solution is to fix the test time using the override env var DATETIME (see README usage).

PR #617, opened 2019-11-07
Updated root url for service-search

nigelatwell:
## Checklist


app/lib/constants.js
```diff
@@ -29,7 +46,7 @@ module.exports = {
     nearby: 'nearby',
     openNearby: 'openNearby',
   },
-  siteRoot: '/find-a-pharmacy',
+  siteRoot: '/service-search/find-a-pharmacy',
```
st3v3nhunt: This should be the only change in this file. The other changes should be reverted.

tomdoughty: I have reverted the linting issues now

tomdoughty:
Changes have been made to the 

> The path in the performance tests will also need updating - `test/performance/pharmacy-finder.jmx`

Thanks Steve. These changes have been made now

app/lib/constants.js
st3v3nhunt: I'm still seeing the changes where the days of week are on separate lines (https://github.com/nhsuk/connecting-to-services/pull/617/files#diff-8216c7e4b1f6e7b1a8d5058460194e55R23) and `description` has still got `"` rather than `'`.

CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+1.18.0 / 2019-11-08
```
st3v3nhunt: I appreciate I'm not in the loop on timings for deployment but I'm a bit surprised this is being released today. If it isn't can the date just be set to `TBC` please?

st3v3nhunt:
Annoyingly the builds have broken due to #609 
It's probably best to wait until Monday and re-run them.

